PR #20, opened 2018-09-25
ci: enable travis

williamcroberts:
Add the CI components to enable travis.

Note: this does not enable coverity or coveralls.


williamcroberts:
Looks like the image is not in the expected state. Closing this to prevent slamming travis with requests as the fork has travis testing too.

Issue #26, opened 2018-10-03
test_pkcs11.c fails with custom TCTI options


robby-cornelissen:
When using the `TPM2_PKCS11_TCTI` environment variable with custom options, e.g.

```bash
export TPM2_PKCS11_TCTI="tabrmd:bus_type=session"
```

the second session test in [`test_pkcs11.c`](https://github.com/tpm2-software/tpm2-pkcs11/blob/9ba93f3b983f83a9fbf7198787e1f00e63129afb/test/unit/test_pkcs11.c) invariably fails, irrespective of the order of the tests.

The problem is caused by [this line](https://github.com/tpm2-software/tpm2-pkcs11/blob/9ba93f3b983f83a9fbf7198787e1f00e63129afb/src/lib/tcti_ldr.c#L192) in `tcti_ldr.c`, which ends up modifying the underlying `TPM2_PKCS11_TCTI` environment variable (see [here](https://stackoverflow.com/a/19746259/3558960) for an explanation), dropping the configuration options.

The offending code looks like this:

```c
char *optstr = getenv (TPM2_PKCS11_TCTI);
// ...
char *split = strchr(optstr, ':');
// ...
split[0] = '\0';
```

Upon execution of the second test, the truncated environment variable will be retrieved, causing the test to fail.

PR #27, opened 2018-10-03
Fixes https://github.com/tpm2-software/tpm2-pkcs11/issues/26

robby-cornelissen:
This should fix the aforementioned issue by copying the `TPM2_PKCS11_TCTI` environment variable so that any modifications made don't impact the original value.

Maybe I've overstepped by doing a little bit of refactoring (clearer variable names and reduced conditional branches), but I leave that to the discretion of the maintainer.

Disclaimer: I'm interested in this project and would like to contribute further, but have never written a line of C before. I tried to do my research and due diligence as best as possible.

### DIFF
--- a/src/lib/tcti_ldr.c
+++ b/src/lib/tcti_ldr.c
@@ -168,49 +168,40 @@ tcti_conf tcti_get_config(void) {
     };
 
     /* no tcti config supplied, get it from env */
-    char *optstr = getenv (TPM2_PKCS11_TCTI);
-    if (!optstr) {
+    char *env_conf_string = getenv(TPM2_PKCS11_TCTI);
+    if (!env_conf_string) {
         /* nothing user supplied, use default */
         return conf;
     }
 
-    char *split = strchr(optstr, ':');
+    char *conf_string = malloc(strlen(env_conf_string) + 1);
+    strcpy(conf_string, env_conf_string);
+
+    char *split = strchr(conf_string, ':');
     if (!split) {
         /* --tcti=device */
-        conf.name = fixup_name(optstr);
-        return conf;
+        conf.name = fixup_name(conf_string);
+    } else {
+        /*
+         * If it has a ":", it could be either one of the following:
+         * case A: --tcti=:               --> default name and default (null) config
+         * case B: --tcti=:/dev/foo       --> default name, custom config
+         * case C: --tcti=device:         --> custom name, default (null) config
+         * case D: --tcti=device:/dev/foo --> custom name, custom config
+         */
+        split[0] = '\0';
+
+        /* Case C, D */
+        if (conf_string[0]) {
+            conf.name = fixup_name(conf_string);
+        }
+
+        /* Case B, D */
+        if (split[1]) {
+            conf.opts = &split[1]; 
+        }
     }
 
-    /*
-     * If it has a ":", it could be either one of the following:
-     * case A: --tcti=:               --> default name and default (null) config
-     * case B: --tcti=:/dev/foo       --> default name, custom config
-     * case C: --tcti=device:         --> custom name, default (null) config
-     * case D: --tcti=device:/dev/foo --> custom name, custom config
-     */
-
-    split[0] = '\0';
-
-    /* Case A */
-    if (!optstr[0] && !split[1]) {
-        return conf;
-    }
-
-    /* Case B */
-    if (!optstr[0]) {
-        conf.opts = &split[1];
-        return conf;
-    }
-
-    /* Case C */
-    if (!split[1]) {
-        conf.name = fixup_name(optstr);
-        return conf;
-    }
-
-    /* Case D */
-    conf.name = fixup_name(optstr);
-    conf.opts = &split[1];
     return conf;
 }
 


src/lib/tcti_ldr.c
```diff
@@ -168,49 +168,40 @@ tcti_conf tcti_get_config(void) {
     };
 
     /* no tcti config supplied, get it from env */
-    char *optstr = getenv (TPM2_PKCS11_TCTI);
-    if (!optstr) {
+    char *env_conf_string = getenv(TPM2_PKCS11_TCTI);
```
williamcroberts: Their is a bug, but your patch doesn't do the bare minimum to fix that bug. Ie 1 patch/commit to fix the bug and another if you want to do refactoring.

I don't see optstr and env_conf_string needing to exist.... you can just "reseat the pointer on optstr" when you need to do the duplication below.

src/lib/tcti_ldr.c
```diff
@@ -168,49 +168,40 @@ tcti_conf tcti_get_config(void) {
...
+    if (!env_conf_string) {
         /* nothing user supplied, use default */
         return conf;
     }
 
-    char *split = strchr(optstr, ':');
+    char *conf_string = malloc(strlen(env_conf_string) + 1);
+    strcpy(conf_string, env_conf_string);
```
williamcroberts: use `strdup()` versus malloc and strcpy

src/lib/tcti_ldr.c
```diff
@@ -168,49 +168,40 @@ tcti_conf tcti_get_config(void) {
...
+         */
+        split[0] = '\0';
 
-    /* Case B */
-    if (!optstr[0]) {
-        conf.opts = &split[1];
-        return conf;
-    }
+        /* Case C, D */
+        if (conf_string[0]) {
```
williamcroberts: You're stated goal here was to reduce branches, but you added a branch and indentation under an else clause to get to the point of a signgle return. This could be all done with else if's but this refactor should be in a different patch.

src/lib/tcti_ldr.c
williamcroberts: You also need to free this. Below you try and get to a single return point, so you'll need a free before the return.

williamcroberts:
FYI That test that introduces this bug will be getting deleted.

src/lib/tcti_ldr.c
```diff
@@ -168,49 +168,40 @@ tcti_conf tcti_get_config(void) {
...
+            conf.name = fixup_name(conf_string);
```
williamcroberts: additionally, fixup name can return static data which can't be free'd. Thus you would need to heap allocate everything.

src/lib/tcti_ldr.c
robby-cornelissen: Thanks for your feedback. I'm definitely not as nimble in C as I am in other programming languages, and I appreciate you taking the time to help me along. I'll try to explain the thought process that went into this patch a little further, so that any flaws in my reasoning become more apparent.

1. Your point about not including unnecessary refactoring in a bug fix is well taken. I didn't introduce the refactoring on a whim, but was trying to address two issues:
    - Branching: I indeed expected to have to `free()` the `conf_string` before the method return (more on that later). Having to do that would have introduced a fairly inelegant label with a number of corresponding `goto` statements, so, instead of doing that, I decided to refactor the branches.
    - Variable naming: The method is trying to populate a `tcti_conf` string with `name` and `opts` fields. Since the `optstr` variable is expected to contain information about both fields, it felt like `confstr` would be a more descriptive name. I then went from `confstr` to `conf_string` based on the coding standards document included in this project (multiple words separated by underscore, no abbreviations except for some exceptions). Rigorously applying the coding standard would have led to `configuration_string` and `environment_configuration_string`, but this would have been inconsistent with the rest of the project (`tcti_conf`, `getenv()` etc.), so I decided against it.
2. `optstr` and `env_conf_string` indeed don't need to coexist, and my patch introduced one as a replacement for the other. As also touched upon above, the renaming came about because I was unsure what name I would use for the copied `optstr`. Both `tmp_optstr` and `optstr_copy` felt a little yuck.
3. I decided on a blanket copy of the environment variable because of [uncertainty](https://stackoverflow.com/questions/19745648/when-getenv-can-alter-content-of-previously-returned-pointer-when-using-glibc) about how subsequent calls to `setenv()` or `putenv()` could potentially affect the value that the pointer obtained from `getenv()` is pointing to.

src/lib/tcti_ldr.c
robby-cornelissen: 1. Agree on using `strdup()` and checking whether memory was successfully allocated.
2. Here is where I'm a little confused and might lack some basic understanding. As noted in my previous comment, it was my intention to free the memory, but calling `free(conf_string)` before returning made all tests fail. My reasoning was that in my implementation, since both fields of `tcti_conf` point to a location in `conf_string`, freeing `conf_string` results in `tcti_conf` no longer having access to its data. This seemed to be confirmed by my tests. Surely I'm misunderstanding something, but I don't quite know what.

src/lib/tcti_ldr.c
robby-cornelissen: Agree on the general principle of separating bug fixes from refactoring.

williamcroberts:
https://github.com/tpm2-software/tpm2-pkcs11/pull/29

PR #32, opened 2018-10-09
fix null pointer dereference in session_table.c

PeterHuewe:
x is calloced a few lines before, so the pointer to ->lock has the value
0 which equals NULL on most systems.
If mutex_create then wants to store the new calloced pthread_mutex_t in
*(x->lock) we get a NULL pointer dereference :/

--> pass the address of x->lock to mutex_create to get the desired
behavior

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>

### DIFF
--- a/src/lib/session_table.c
+++ b/src/lib/session_table.c
@@ -26,7 +26,7 @@ CK_RV session_table_new(session_table **t) {
         return CKR_HOST_MEMORY;
     }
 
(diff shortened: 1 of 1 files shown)


williamcroberts:
Fixed: https://github.com/tpm2-software/tpm2-pkcs11/pull/31/commits/d025e712615409bb561f0d9e82698dd79fc8c13f

PR #34, opened 2018-10-10
Added more test cases to integration tests and deleted test_pkcs11.c

kbindlis:
Unit tests from test_pkcs11.c have been moved to integration tests as these tests needed the integration framework to be in place. Al the tests are currently passing.


williamcroberts:
Passes on GCC fails on Clang, likely memory leak or other memory issue

kbindlis:
Removed the test case with the memory leak. Will be creating a new test case after some more investigation.

PR #36, opened 2018-10-11
Apply correct padding

PeterHuewe:
Most of the PKCS#11 strings MUST NOT be NULL terminated and MUST be padded with ' '.
This patchset fixes a few places 


PeterHuewe:
First two fixes for #37 - more occurences are available.

williamcroberts:
We should probably allocate strings 1 bigger (ie label as size 33), do the space pad but keep the terminating zero. If applications shouldn't be looking for a 0 they should still function fine, but someone using a C str routine won't end up in never never land.

PR #39, opened 2018-10-12
Add support for key labels

PeterHuewe:
Some applications want to retrieve the key by a key label.

### DIFF
--- a/tools/tpm2_ptool.py
+++ b/tools/tpm2_ptool.py
@@ -80,6 +80,7 @@ CKK_AES = 0x1f
 CKM_RSA_PKCS_OAEP = 9
 CKM_AES_CBC = 0x1082
 
+CKA_LABEL = 0x3
 CKA_ID = 0x102
 CKA_MODULUS = 0x120
 CKA_PUBLIC_EXPONENT = 0x122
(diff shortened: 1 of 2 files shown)


tools/tpm2_ptool.py
```diff
@@ -924,6 +925,10 @@ def generate_options(self, group_parser):
             help='The type of the key.\n',
             choices=[ 'rsa1024', 'rsa2048', 'aes128', 'aes256' ],
             required=True)
+        group_parser.add_argument(
+            '--keylabel',
```
williamcroberts: `--key-label`

tools/tpm2_ptool.py
```diff
@@ -924,6 +925,10 @@ def generate_options(self, group_parser):
...
+            help='The key label to identify the key. Defaults to testkey\n',
+            default="testkey")
```
williamcroberts: Then you'll have duplicate labels. Why not use the primary key integer from the db as a string?

williamcroberts: perhaps output the label in addkey, which if you use the primary key for the object in the db, is already correct. So it just needs minor logic tweak.

tools/tpm2_ptool.py
```diff
@@ -1059,6 +1065,7 @@ def __call__(self, args):
 
                 if alg.startswith('rsa'):
                     attrs = [
+                        {  CKA_LABEL           : keylabel        },
```
williamcroberts: label would apply to all objects, not just RSA

williamcroberts:
#46 continues this work and resolves the issues mentioned as well as adds a test for it.

PR #40, opened 2018-10-12
C_FindObjectsInit: Return all objects if ulCount is 0

PeterHuewe:
According to 5.7 PKCS#11 Spec C_FindObjectsInit must return all objects
if ulCount is 0.
If ulCount is 0, templ MAY be NULL (as it is unused), so we have to move
the check.


williamcroberts:
@PeterHuewe I'll add tests for your 2 PRs today and get them merged.

williamcroberts:
I moved this onto #46 as the test hunk would conflict

PeterHuewe:
Ok works for me

PR #46, opened 2018-10-16
add support for key labels and empty template find object calls

williamcroberts:
Peter Huewe (1):
      C_FindObjectsInit: Return all objects if ulCount is 0

William Roberts (2):
      tpm2_ptool.py: fix addtertiary variable name
      db: add support for key labels

### DIFF
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -232,6 +232,16 @@ static bool parse_attrs(const char *key, const char *value, size_t index, void *
             return false;
         }
     } break;
+
+    case CKA_LABEL: {
+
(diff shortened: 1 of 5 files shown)


PeterHuewe:
*Ass* Support? Maybe you want to correct that typo :)

williamcroberts:
> _Ass_ Support? Maybe you want to correct that typo :)

The major design flaw of the qwerty keyboard

williamcroberts:
> _There_

Got those fixups for both patches

PeterHuewe:
Looks good to me.

PR #47, opened 2018-10-17
Fix some locking

williamcroberts:
William Roberts (2):
      mutex: fix some locking issues
      mutex: enable lock/unlock logs


stefanberger:
Works for me :-) :+1:

PR #48, opened 2018-10-17
C_GetAttributeInfo: fix deadlock

williamcroberts:
If a buffer is too small, it triggers an immediate return
without unlocking the session_ctx lock. Set the return
value in rv and goto out instead.


stefanberger:
:+1:

Issue #5, opened 2018-09-21
Session State


williamcroberts:
Session State is currently not global and it should be. Once a C_Login occurs, all sessions are logged in. Correct this in the implementation and test.

robby-cornelissen:
I'm looking into trying to take this on, but have the following question: when you say "global", what do you mean exactly?

In the Cryptoki specification, I find the following:

> Call **C_Login** to log the user into the token.  Since all sessions an application has with a token have a shared login state, **C_Login** only needs to be called for one of the sessions.

If I read that correctly, that would mean that a minimal implementation would have to:

1. Add the `application` to the `session_ctx` struct.
2. In `session_login()`, after `session_ctx_login()` returns, walk through the `session_table` and update the state for all other sessions with the same application/token.
3. Do the same for `session_logout()`.

Am I on the right track here?

williamcroberts:
2 and 3 seem correct, I'm not sure why you need 1.

PR #49, opened 2018-10-18
Fixes https://github.com/tpm2-software/tpm2-pkcs11/issues/5

robby-cornelissen:
Should fix the linked issue. I'm a little bit unsure about the locking semantics. I currently lock the session table to update all the session contexts, but it might be safer to lock the individual session contexts as well.

### DIFF
--- a/src/lib/session_ctx.h
+++ b/src/lib/session_ctx.h
@@ -55,7 +55,9 @@ tpm_ctx *session_ctx_get_tpm_ctx(session_ctx *ctx);
 
 token *session_ctx_get_token(session_ctx *ctx);
 
+void session_ctx_state_set(session_ctx *ctx, session_ctx_state state);
 session_ctx_state session_ctx_state_get(session_ctx *ctx);
+
 
 static inline bool session_is_rw(session_ctx *ctx) {
 
--- a/src/lib/session_table.c
+++ b/src/lib/session_table.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2018, Intel Corporation
  * All rights reserved.
  */
+#include <stdio.h>
 #include <stdlib.h>
 
 #include "assert.h"
@@ -10,6 +11,7 @@
 #include "pkcs11.h"
 #include "session_ctx.h"
 #include "session_table.h"
+#include "token.h"
 #include "utils.h"
 
 struct session_table {
@@ -160,4 +162,20 @@ unlock:
     session_table_unlock(t);
 
     return ctx;
+}
+
+void session_table_update_ctx_state(session_table *t, token *tok, session_ctx_state state) {
+    session_table_lock(t);
+
+    unsigned i;
+    for (i=0; i < ARRAY_LEN(t->table); i++) {
+        CK_SESSION_HANDLE handle = i;
+        session_ctx *ctx = *session_table_lookup_unlocked(t, handle);
+
+        if (ctx && session_ctx_get_token(ctx) == tok) {
+            session_ctx_state_set(ctx, state);
+        }
+    }
+
+    session_table_unlock(t);
 }
--- a/src/lib/session.c
+++ b/src/lib/session.c
@@ -138,6 +138,13 @@ CK_RV session_login (CK_SESSION_HANDLE session, CK_USER_TYPE user_type,
     twist_free(tpin);
     session_ctx_unlock(ctx);
 
+    if (rv == CKR_OK) {
+        token *tok = session_ctx_get_token(ctx);
+        session_ctx_state state = session_ctx_state_get(ctx);
+
+        session_table_update_ctx_state(global.s_table, tok, state);
+    }
+
     return rv;
 }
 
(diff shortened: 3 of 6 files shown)


src/lib/session_ctx.h
```diff
@@ -55,8 +55,10 @@ tpm_ctx *session_ctx_get_tpm_ctx(session_ctx *ctx);
 
 token *session_ctx_get_token(session_ctx *ctx);
 
+void session_ctx_state_set(session_ctx *ctx, session_ctx_state state);
 session_ctx_state session_ctx_state_get(session_ctx *ctx);
 
+
```
williamcroberts: spurious white-space changes in patches are frowned apon

src/lib/session_table.c
```diff
@@ -3,13 +3,15 @@
  * Copyright (c) 2018, Intel Corporation
  * All rights reserved.
  */
+#include <stdio.h>
```
williamcroberts: why did you add stdio?

robby-cornelissen:
I really must be misunderstanding something here. From the spec, I concluded that the login state is supposed to be shared on the application/token level.

By moving the state to the session table, it would be made effectively global, which precludes the possibility of an application opening sessions with multiple tokens. While this is probably not the most common use case, nothing in the spec seems to prohibit it.

Or were you suggesting to maintain some sort of token/state map in the session table?

src/lib/session_table.c
robby-cornelissen: Sorry about that. Leftover from some debug logging.

williamcroberts:
An application will link the library where the state is stored in the token state, thus for N applications/procceses their will be N states.

With that said, when login to a smart card that login is good until:
The physical token resets the I'm logged in state. For piv cards this occurs after every command or the token is removed and unpowered.

williamcroberts:
Oh I think when you said application you meant card side. We can't do that. The shared login concept will unfortunately be limited to host side process scope

williamcroberts:
And actually your code is effectively a complicated global token state flag, you don't actually get th properties you describe.

robby-cornelissen:
I think we're talking past each other here. I have understood that the library state is not shared across applications. Therefore, In my patch, I did not put any logic in place to keep track of the `pApplication` callback pointers (which I at first mistook for application identifiers).

What I'm still unclear about is whether or not an application can access multiple tokens during its lifetime, since nothing in the specification seems to preclude that. If an application is indeed able to access multiple tokens, then login state should not be shared across **tokens** (as per the specification).

I don't understand what you mean when you say that my code is "effectively a complicated global token state flag", because the logic takes the actual token into account when propagating state, and I have put unit tests in place to verify this.

To summarize, your point of view seems to be:

> [...] for N applications/procceses there will be N states [...]

While mine is:

> For N applications/processes and M tokens, there will be up to N*M states.

williamcroberts:
> -----Original Message-----
> From: Robby Cornelissen [mailto:notifications@github.com]
> Sent: Monday, October 22, 2018 6:59 PM
> To: tpm2-software/tpm2-pkcs11 <tpm2-pkcs11@noreply.github.com>
> Cc: Roberts, William C <william.c.roberts@intel.com>; Comment
> <comment@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-pkcs11] Fixes https://github.com/tpm2-
> software/tpm2-pkcs11/issues/5 (#49)
> 
> I think we're talking past each other here. I have understood that the library state
> is not shared across applications. Therefore, In my patch, I did not put any logic in
> place to keep track of the pApplication callback pointers (which I at first mistook
> for application identifiers).
> 
> What I'm still unclear about is whether or not an application can access multiple
> tokens during its lifetime, since nothing in the specification seems to preclude
> that. If an application is indeed able to access multiple tokens, then login state
> should not be shared across tokens (as per the specification).

Ahh I see our point of confusion. When I say global token flag, I really mean a 
Field in the token specific data structure. That structure is specific to that token
And would only affect sessions on that token, not across other tokens.

Does that make more sense?

williamcroberts:
I went through and read https://www.cryptsoft.com/pkcs11doc/STANDARD/pkcs-11.pdf around session handling state. looks like state is per-session and login events update existing. I'm unsure if the spec expects new sessions will pick up the login state.

In this model, if Thread 1 (T1) is the one updating the session states, and T2 is using it's session state, T2 could easily miss a login event. Granted T1's C_Login() wouldn't have returned and a "login event" shouldn't be considered propagated until C_Login() returns. Thus T2 could have an i'm not logged in error because T1 hasn't gotten to change it's state yet. Grabbing those session_ctx locks and iterating could be slow.

I almost wonder if:
1. We keep an atomic or mutex protected variable in the token structure to indicate a user is logged in.
2. Each thread updates their own session state, ie every function that takes session checks the atomic 
    and transitions.

williamcroberts:
I fixed those two issues and moved it to: https://github.com/tpm2-software/tpm2-pkcs11/pull/71

I consider any of those comments above as performance improvements.

PR #57, opened 2018-10-23
tpm: move to esapi api

williamcroberts:
use the ESAPI api so sessions and HMAC auths becomes easier
in the future.


williamcroberts:
Their is definitely some weirdness to the tpm_api in that it expects an always write able "handle" so that when setting the auth for the object, if the handle is not a ESYS_TR object we can detect and convert it.

I don't think that was the best decision, considering that it's only  the primary handle that is a raw tpm handle that needs to be converted to an ESYS_TR.

I'm thinking an explicit tpm_convert_handle() call for the primary object would be a better approach.

PR #58, opened 2018-10-23
Configure and twist: fix some build issues

vtcao297:
-fix cmocka dependency checks required even when --enable-unit is disabled
-fix twist compile error when UNIT_TESTING is not defined

Signed-off-by: Vincent Cao <vincent.t.cao@intel.com>


williamcroberts:
The changes look good but it fails travis.

vtcao297:
travis issue should be fixed now

PR #66, opened 2018-11-01
New PR with all the review feedback incorporated.

ichetty:
1. Fixed the issue with Sign operation
2. Added new import commandlet
3. Made some changes to tpm2_ptool.py based on testing on discrete tpm


ichetty:
The CI build will fail because the integration test depends on another PR in the tpm2-tools.

ichetty:
New PR with all the review feedback addressed.

PR #71, opened 2018-11-06
fixes https://github.com/tpm2-software/tpm2-pkcs11/issues/5

williamcroberts:

### DIFF
--- a/src/lib/session_table.c
+++ b/src/lib/session_table.c
@@ -10,6 +10,7 @@
 #include "pkcs11.h"
 #include "session_ctx.h"
 #include "session_table.h"
+#include "token.h"
 #include "utils.h"
 
 struct session_table {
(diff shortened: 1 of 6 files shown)


src/lib/session_table.c
```diff
@@ -161,3 +162,19 @@ session_ctx *session_table_lookup(session_table *t, CK_SESSION_HANDLE handle) {
...
+
+void session_table_update_ctx_state(session_table *t, token *tok, session_ctx_state state) {
+    session_table_lock(t);
+
+    unsigned i;
+    for (i=0; i < ARRAY_LEN(t->table); i++) {
+        CK_SESSION_HANDLE handle = i;
+        session_ctx *ctx = *session_table_lookup_unlocked(t, handle);
+
+        if (ctx && session_ctx_get_token(ctx) == tok) {
```
williamcroberts: need to hold session_lock here.

Also, theirs a major data structure flaw. The session table should be per token. So what we need to do is initialize a session table for each token. The only data structure that should be global is the token list. Everything from there should be in the token. Ill update all of this and bring this patch to merge.

williamcroberts: However, we can't just issue 2 of the same value session handles, as we'll lose what token they refer to. I think we need to say you can support 256 tokens in the design, and `~(CK_SESSION_HANDLE(0)) >> 8` number of sessions

PR #73, opened 2018-11-08
Session table token scope

williamcroberts:
* move the session table to token scope
* test that C-CloseAllSessions() stays within token scope

Fixes: #74
Fixes: #5


williamcroberts:
This change will lay the foundation for:
1. Getting session state flags and states sorted (ie drop our internal types and use the CK_STATE and CK_FLAGS types.
2. Start tracking if a SO session or USER sessions are active and deal with errors.
3. Login all current and future sessions

See:
  - https://www.cryptsoft.com/pkcs11doc/STANDARD/pkcs-11.pdf
  - https://www.cryptsoft.com/pkcs11doc/v220/group__SEC__6__7__2__READ__WRITE__SESSION__STATES.html

PR #76, opened 2018-11-12
db: fix stat LOGV message

williamcroberts:
When not able to stat the db, the LOGV message was reporting
the base_path, which is either PKCS11_STORE_ENV_VAR or HOME
environment variables.

Use the full constructed path as constructed by snprintf().


williamcroberts:
@ichetty

PR #82, opened 2018-11-15
pkcs11.h: structure packing: dont do it on linux

williamcroberts:
Despite section 2.1 of:
  - http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959683

Stating:
  Cryptoki structures are packed to occupy as little space as is possible.
  Cryptoki structures SHALL be packed with 1-byte alignment.

Linux seems to not pack the structures. On a 64 bit machine, for instance
C_Initialize in libp11-kit's pkcs11 header file is at offset 8, versus offset
2. The only thing in-front of the C_Initialize function pointer is 2 char
inline version structure.

Stop packing on Linux, detect Windows and allow a FORCE_PACK option via
configure --enable-force-pack.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -97,6 +97,14 @@ AC_ARG_ENABLE([dlclose],
   [AC_DEFINE([DISABLE_DLCLOSE], [1])]
  )
 
+AC_ARG_ENABLE([pack],
+            [AS_HELP_STRING([--enable-pack=]@<:@yes/no@:>@,
+                            [Pack the structures. (default is no, except on Windows, where it defaults to packing)])],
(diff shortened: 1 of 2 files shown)


williamcroberts:
@PeterHuewe does this fix at least one of your issues?

williamcroberts:
Darn stale m4 cache.... re-spinning

PeterHuewe:
Works for me and looks good to me +1
Reviewed-by: Peter Huewe <peterhuewe@gmx.de>

Issue #85, opened 2018-11-19
Keygen not implemented, returns dummy handles


PeterHuewe:
The current key_gen returns just two dummy handles, which would confuse an application using this functionality.
Either we should return CKR_FUNCTION_NOT_SUPPORTED for now, or better implement / re-enable the code.

PR #87, opened 2018-11-27
C_GenerateKeyPair: return CKR_FUNCTION_NOT_SUPPORTED

williamcroberts:
It's not implemented, just stubbed in and un-tested. Just
return CKR_FUNCTION_NOT_SUPPORTED for now so applications
don't get confused.

Fixes: #85


williamcroberts:
```
../src/lib/key.c:62:5: warning: Value stored to 'rv' is never read
```

PR #100, opened 2018-12-06
tpm2-pkcs11 authorization model to restrict SO operations to PIN changes and USER operations to create and operate on end user objects

idesai:


williamcroberts:
Wouldn't using "" (empty string) be the empty password, did you try that?

williamcroberts:
Granted handling "" and NULL does provide a layer of robustness to the code. You need to modify test so this is tested.

tools/tpm2_ptool.py
```diff
@@ -133,13 +133,18 @@ def __init__(self, tmp, final):
...
-                    '-o', ctx]
+                        '-o', ctx]
+
+        if objauth and len(objauth) > 0:
+            cmd.extend(['-p', 'hex:%s' % objauth.decode()])
 
         if ownerauth and len(ownerauth) > 0:
             cmd.extend(['-P', ownerauth])
 
+        print ("%s",cmd)
```
idesai: stray debug comment

williamcroberts:
This looks like an overly complicated way to solve this. Id like to see how you expect to call tpm2_ptool.py commands.

src/lib/token.c
```diff
@@ -228,9 +228,11 @@ CK_RV token_login(token *tok, twist pin, CK_USER_TYPE user) {
     twist pobjsalt = user == CKU_USER ? tok->userpobjauthkeysalt : tok->sopobjauthkeysalt;
     twist pobjauth = user == CKU_USER ? tok->userpobjauth : tok->sopobjauth;
 
-    dpobjauth = decrypt(pin, pobjsalt, pobjiters, pobjauth);
-    if (!dpobjauth) {
-        return CKR_PIN_INCORRECT;
+    if(pobjiters) {
```
williamcroberts: Wouldn't this be the encrypted null byte and then that get's passed to the tools (empty string) and it should work. All the functionality that is broken Is likely all in the tpm2_ptool.py script.

Checking pobjiters is not correct here. Also style on parens.

src/lib/token.c
```diff
@@ -263,6 +265,7 @@ CK_RV token_login(token *tok, twist pin, CK_USER_TYPE user) {
     wrappingobject *wobj = &tok->wrappingobject;
     twist wobjauth = tpm_unseal(tpm, sealobjhandle, sealobjauth);
     if (!wobjauth) {
+        rv = CKR_PIN_INCORRECT;
```
williamcroberts: This looks like a bug and should be split out into a separate patch and can be uploaded independently. I would take that as is.

tools/tpm2_ptool.py
```diff
@@ -871,7 +883,9 @@ def do_token_init(db, path, args):
         # Verify pid is in db
         pobject = db.getprimary(pid)
 
-        pobjauthhash = AddTokenCommand.verify_pobjpin(pobject, pobjpin)
```
williamcroberts: an empty pobjpin should be passing "" here, is it?

src/lib/token.c
idesai: Agreed. Creating another PR for this one.

src/lib/token.c
idesai: @williamcroberts you are right. Checking pobjiters wasn't necessary.

tools/tpm2_ptool.py
idesai: @williamcroberts right again.

idesai:
> This looks like an overly complicated way to solve this. Id like to see how you expect to call tpm2_ptool.py commands.

@will Added tests for passwordless primary objects in create_pkcs_store.sh, please take a look.

will:
fuck yeah, tests are great

test/integration/scripts/create_pkcs_store.sh
```diff
@@ -97,6 +97,15 @@ echo "importing rsa2048 key under token 'import-keys'"
...
 tpm2_ptool.py import --privkey='private.pem' --algorithm=rsa --key-label="imported_key" --label="import-keys" --userpin=anotheruserpin --path=$TPM2_PKCS11_STORE
 
+#Passwordless Primary
+tpm2_ptool.py init --path=$TPM2_PKCS11_STORE
+##Add token
+tpm2_ptool.py addtoken --pid=3 --sopin=mysopin --userpin=myuserpin --label=pwless --path $TPM2_PKCS11_STORE
+###Add aes key under token
+tpm2_ptool.py addkey --algorithm=aes256 --label=pwless --userpin=myuserpin --path=$TPM2_PKCS11_STORE
+##Verify setup
+tpm2_ptool.py verify --sopin=mysopin --userpin=myuserpin --label=pwless --path=$TPM2_PKCS11_STORE
```
williamcroberts: A bunch of tests are failing, probably hard-code the number of tokens expected. Please ensure make check passes.

idesai: Thanks for the pointer

idesai:
@williamcroberts, let's not merge just yet :) More commits coming. Also have to rebase after C_initpin and C_setpin. Thanks.

williamcroberts:
> -----Original Message-----
> From: Imran Desai [mailto:notifications@github.com]
> Sent: Monday, January 28, 2019 12:30 PM
> To: tpm2-software/tpm2-pkcs11 <tpm2-pkcs11@noreply.github.com>
> Cc: Roberts, William C <william.c.roberts@intel.com>; Mention
> <mention@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-pkcs11] tpm2-pkcs11 authorization model to
> restrict SO operations to PIN changes and USER operations to create and operate
> on end user objects (#100)
> 
> @williamcroberts <https://github.com/williamcroberts>  should this be a
> separate branch prior to merge?

Everything comes in as a separate branch from your repository. I'd prefer
A rebase workflow.

PR #106, opened 2018-12-26
 test: fix race condition in create_pkcs_store.sh

diabonas:
Multiple integration tests can be run in parallel, but only one file `primary.ctx` is used for all of them in [`create_pkcs_store.sh`](https://github.com/tpm2-software/tpm2-pkcs11/blob/5be9faf7399b18e1f4f363af3d455e08eb17aba4/test/integration/scripts/create_pkcs_store.sh#L59). On my system, this leads to errors of the form
```
ERROR: Tss2_Sys_ContextLoad(0x1DF) - tpm:parameter(1):integrity check failed
ERROR: Could not load object, got: "primary.ctx"
ERROR: Unable to run tpm2_evictcontrol
```
when multiple tests try to write to that file at once. This is easily fixed by using separate files per test in the `$TPM2_PKCS11_STORE` directory.

In unrelated news, remove the instructions to rebuild `twist.c` because this is not necessary any more since #96 switched to using `config.h`.

### DIFF
--- a/test/integration/scripts/create_pkcs_store.sh
+++ b/test/integration/scripts/create_pkcs_store.sh
@@ -56,8 +56,8 @@ set -e
 tpm2_ptool.py init --pobj-pin=mypobjpin --path=$TPM2_PKCS11_STORE
 
 # Test the existing primary object init functionality
-tpm2_createprimary -p foopass -o primary.ctx -g sha256 -G rsa
-handle=`tpm2_evictcontrol -a o -c primary.ctx | cut -d\: -f2-2 | sed 's/^ *//g'`
+tpm2_createprimary -p foopass -o $TPM2_PKCS11_STORE/primary.ctx -g sha256 -G rsa
+handle=`tpm2_evictcontrol -a o -c $TPM2_PKCS11_STORE/primary.ctx | cut -d\: -f2-2 | sed 's/^ *//g'`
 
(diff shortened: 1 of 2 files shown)


test/integration/scripts/create_pkcs_store.sh
```diff
@@ -56,8 +56,9 @@ set -e
 tpm2_ptool.py init --pobj-pin=mypobjpin --path=$TPM2_PKCS11_STORE
 
 # Test the existing primary object init functionality
-tpm2_createprimary -p foopass -o primary.ctx -g sha256 -G rsa
-handle=`tpm2_evictcontrol -a o -c primary.ctx | cut -d\: -f2-2 | sed 's/^ *//g'`
+primary_context=`mktemp`
+tpm2_createprimary -p foopass -o "$primary_context" -g sha256 -G rsa
+handle=`tpm2_evictcontrol -a o -c "$primary_context" | cut -d\: -f2-2 | sed 's/^ *//g'`
```
williamcroberts: we should probably do this as $TPM2_PKCS11_STORE/primary.ctx
so an rm -rf of TPM2_PKCS11_STORE deletes everything.

diabonas: Thanks, that makes more sense indeed, implemented!

PR #107, opened 2019-01-03
token.c: Return pk11 defined error-condition: incorrect-pin  when unsealing fails

idesai:


src/lib/token.c
```diff
@@ -263,6 +263,7 @@ CK_RV token_login(token *tok, twist pin, CK_USER_TYPE user) {
     wrappingobject *wobj = &tok->wrappingobject;
     twist wobjauth = tpm_unseal(tpm, sealobjhandle, sealobjauth);
     if (!wobjauth) {
+        return CKR_PIN_INCORRECT;
```
williamcroberts:
```C
rv = CKR_PIN_INCORRECT;
goto error;
```

PR #114, opened 2019-01-24
tpm2_ptool: remove sym-support conditional blocks

williamcroberts:
The code was littered with checking if the tpm config supported symmetric
encryption, and if so wrap or unwrap object authorizations differently.

This caused two issues:
1. the code was harder to follow
2. the verify commandlet not to be updated when the SW wrapping
   support was added and thus was broken.

This fixes these two issues.


tools/tpm2_ptool.py
```diff
@@ -1166,40 +1196,29 @@ def new_key_init(label, sopin, userpin, db, tpm2):
...
 
         # decrypt sobj auth with wrapping
-        sobjauth = sobj['objauth'];
-        if sym_support:
-            sobjauth = binascii.unhexlify(sobjauth)
-            sobjauth = tpm2.decrypt(wrappingctx, wrappingauth, sobjauth)
-        else:
-            c = AESCipher(binascii.unhexlify(wrappingauth))
-            sobjauth = c.decrypt(sobjauth)
+        encsobjauth = sobj['objauth'];
```
williamcroberts: semi-colon

PR #115, opened 2019-01-25
tpm2_ptool.py: implement changepin

williamcroberts:
Implement the "changepin" functionality. This allows a user
or so to change their pin IF they know the old pin. It does
so by updating the Software Wrapping key for the primary
object auth value, as well as updating the auth value for
the coresponding users sealing object via tpm2_changeauth.

It re-populated the coresponding db fields with the new
information.


williamcroberts:
Depends on tpm2-tools change:
https://github.com/tpm2-software/tpm2-tools/pull/1290

PR #124, opened 2019-02-01
db: store tpm pub/priv data in db

williamcroberts:
** This will require a db delete and restore **

Having the public and private tpm data in files with paths in the DB itself
makes creating new objects that much more difficult as one needs to deal with
the myriad of filesystem issues. It would be much simpler to just store the
blob data in the DB, and should be faster:
  - https://www.sqlite.org/intern-v-extern-blob.html

### DIFF
--- a/tools/tpm2_pkcs11/tpm2.py
+++ b/tools/tpm2_pkcs11/tpm2.py
@@ -4,17 +4,16 @@ import os
 import shutil
 import sys
 import tempfile
+from tempfile import NamedTemporaryFile
 import uuid
 import yaml
 
 from subprocess import Popen, PIPE
 
-
 class Tpm2(object):
 
-    def __init__(self, tmp, final):
+    def __init__(self, tmp):
         self._tmp = tmp
-        self._final = final
 
     def createprimary(self, ownerauth, objauth):
         ctx = os.path.join(self._tmp, "context.out")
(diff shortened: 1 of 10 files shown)


tools/tpm2_pkcs11/tpm2.py
```diff
@@ -53,6 +52,18 @@ def evictcontrol(ownerauth, ctx):
 
     def load(self, pctx, pauth, priv, pub):
 
+        if isinstance(priv, buffer):
```
diabonas: The `buffer` type doesn't exist any more in Python 3 and has been replaced by [`bytes` in `sqlite3`]( https://docs.python.org/3/library/sqlite3.html#sqlite-and-python-types). Using
```
if not isinstance(priv, str):
```
works in Python 2 and 3.

tools/tpm2_pkcs11/tpm2.py
```diff
@@ -53,6 +52,18 @@ def evictcontrol(ownerauth, ctx):
...
+            sealprivf = NamedTemporaryFile()
+            sealprivf.write(priv)
+            sealprivf.flush()
+            priv = sealprivf.name
+
+        if isinstance(pub, buffer):
```
diabonas: Same as above

williamcroberts: @diabonas so glad you added the PYTHON3 tests into the CI system. I'm liking that were failing up-front!

Issue #142, opened 2019-02-20
move opdata back into session_ctx


williamcroberts:
Undo the changes for #6 as it should be session based.

PR #155, opened 2019-03-04
session_ctx: move opdata back from token scope

williamcroberts:
Operation state for things like CKR_OPERATION_ACTIVE should be
session scope, not token scope. Correct this by moving opdata
managment back into the session_ctx structure. Also, since some
memory leaks occur if C_Logout() and C_CloseSession() are not
called, and entangled in the opdata managment, fix those leaks
as well.

Fixes: #142


williamcroberts:
@ichetty this should fix the operation active state issue you noticed.

PR #162, opened 2019-03-10
configure.ac: “test ==” is not portable, use “test =” instead

dilyanpalauzov:


configure.ac
```diff
@@ -301,7 +301,7 @@ AX_CHECK_COMPILE_FLAG([-fdata-sections], [strip+="y"])
 AX_CHECK_COMPILE_FLAG([-ffunction-sections], [strip+="y"])
```
diabonas: While you are at it, could you also fix the non-portable `+=` in ll. 300-302 by using `strip="${strip}y"` (cf. https://github.com/tpm2-software/tpm2-tools/pull/1277) instead?

PR #181, opened 2019-03-26
Return correct values for empty attributes in C_GetAttributeValue

PeterHuewe:
See Spec C_GetAttributeValue
"2.     Otherwise, if the specified value for the object is invalid (the
object does not possess such an attribute), then the ulValueLen field in
that triple is modified to hold the value CK_UNAVAILABLE_INFORMATION.

"If case 2 applies to any of the requested attributes, then the call
should return the value CKR_ATTRIBUTE_TYPE_INVALID."

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -118,6 +118,7 @@ TESTS = $(check_PROGRAMS)
 check_PROGRAMS += \
     test/integration/pkcs-find-objects.int \
     test/integration/pkcs-get-mechanism.int \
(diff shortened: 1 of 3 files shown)


PeterHuewe:
I will write some tests for this next week.

PeterHuewe:
currently working on the test - I'll rebase in this step.

PeterHuewe:
@williamcroberts  please review.

PR #184, opened 2019-03-28
Include config.h for --with-storedir to work

PeterHuewe:
config.h needs to be included in order for configure --with-storedir to work.

Fixes: #29b2536f ("db: add search logic")

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>

### DIFF
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -18,6 +18,7 @@
 
 #include <sqlite3.h>
 
+#include "config.h"
 #include "db.h"
 #include "log.h"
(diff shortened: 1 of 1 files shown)


williamcroberts:
our CI might be down, do don't worry too much if this fails

diabonas:
The Docker image is [updated](https://github.com/tpm2-software/tpm2-software-container/pull/3), so rebasing this PR (or directly restarting the build) should hopefully cause the CI to pass.

PR #185, opened 2019-04-02
Fix CodeCov.io Codecoverage

PeterHuewe:
The current codecov.io code coverage job in travis fails,
since the testing is done within a docker container and the
outside job cannot access the coverage reports.

https://docs.codecov.io/docs/testing-with-docker provides two solutions
for this - either copy the files via a shared folder, or just run
codecov within the container.

We now run it within the container, for this we need to install curl
first.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -8,7 +8,7 @@
 set -e
 
 source $TRAVIS_BUILD_DIR/.ci/docker-prelude.sh
-
+apt-get install -y curl
 if [ -d build ]; then
   rm -rf build
 fi
@@ -83,6 +83,10 @@ fi
 ../configure --enable-unit --enable-integration --enable-esapi-session-manage-flags $config_flags
 make -j$(nproc)
 make -j$(nproc) check
+
+if [ "$CC" == "gcc" ] && [ "$PYTHON_INTERPRETER" == "python3" ]; then
+    bash <(curl -s https://codecov.io/bash)
+fi
 
 popd
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,3 @@ script:
   - ./.ci/travis.run
 after_failure:
   - cat build/test-suite.log
-after_success:
-  - |
-    if [ "$CC" == "gcc" ] && [ "$PYTHON_INTERPRETER" == "python3" ]; then
-        bash <(curl -s https://codecov.io/bash)
-    fi
--- a/.ci/travis.run
+++ b/.ci/travis.run
@@ -19,7 +19,8 @@ if [ "$TRAVIS_BRANCH" != "coverity_scan" ]; then
   #
   # Execute the build and test procedure by running .ci/docker.run
   #
-  docker run --env-file .ci/docker.env \
+  ci_env=`bash <(curl -s https://codecov.io/env)`
+  docker run $ci_env --env-file .ci/docker.env \
     -v `pwd`:/workspace/tpm2-pkcs11 tpm2software/tpm2-tss \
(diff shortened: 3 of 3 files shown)


.ci/docker.run
```diff
@@ -8,7 +8,7 @@
 set -e
 
 source $TRAVIS_BUILD_DIR/.ci/docker-prelude.sh
-
+apt-get install -y curl
```
williamcroberts: This should go over to our container image: https://github.com/tpm2-software/tpm2-software-container

.travis.yml
```diff
@@ -13,8 +13,3 @@ script:
   - ./.ci/travis.run
 after_failure:
   - cat build/test-suite.log
-after_success:
```
williamcroberts: Is this really the problem? If I can access build/test-suite.log then why can't I access the other data? I share this folder with the docker container so that when it's done, I can get any data I need.

If this is the problem, then why do I have any coverage data uploaded?

.ci/docker.run
PeterHuewe: Done.

.travis.yml
PeterHuewe: I'm not 100% sure - but by putting it in the docker file, as indicated in the documentation it works.

The only test coverage you get is by twist which is a unit test (compared to integration) - which is indeed strange.

By examining the output of the two solutions, 
travis.yml version outputs
```
==> Travis CI detected.
    project root: .
--> token set from env
    Yaml found at: codecov.yml
==> Running gcov in . (disable via -X gcov)
```

docker version outputs
```
==> Travis CI detected.
    project root: /workspace/tpm2-pkcs11
    Yaml found at: codecov.yml
==> Running gcov in /workspace/tpm2-pkcs11 (disable via -X gcov)
```

Maybe it the path needs to be specified absolutely or something similar.

The folder is shared, as in the documentation, but in a little bit different manner
`docker run -v "$PWD/shared:/shared" ...`
vs
`docker run -v `pwd`:/workspace/tpm2-pkcs11`

So the directories are mapped a little bit differently - maybe that's the reason.

PeterHuewe: The other problem, when running codecov from travis and not from docker, is that the version of gcc and gcov are mismatching.
This is where this version **'504*', prefer '408*'** comes from.

The docker environment uses "gcc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609"
The travis environment used gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

Maybe we also should consider updating our travis env (and docker) to xenial. (18.04)
Trusty is 5 years old now....

williamcroberts: Yep agree. I'd like us to have multiple images, and cron-job master on a few different distros and versions and then check PR's against something like the newest stable Ubuntu.

williamcroberts:
> I'm not 100% sure - but by putting it in the docker file, as indicated in the documentation it works.

I'm willing to try your changes.

> The only test coverage you get is by twist which is a unit test (compared to integration) - which is indeed strange.

At one point in time I had all the data and the coverage was around 75%, then it jumped to 95% and stayed flat. I am also seeing no new uploads hit the codecov account:
https://codecov.io/gh/tpm2-software/tpm2-pkcs11/tree/19a2381a7e3ab7c1a5bb9b0170fb69f0fd1c8e55

Which yours is showing:
https://codecov.io/gh/tpm2-software/tpm2-pkcs11/pull/185

This stuff is a PITA to debug. I guess just get the container updated and when I see the image update ill merge this. Thanks again.

PR #187, opened 2019-04-02
Misc fixes on python part

maage:
setup.py to have long_description field and now twine check is happy

And then move tmpfile creation after error handling so they are not left dangling.


williamcroberts:
@maage thanks.

PR #191, opened 2019-04-05
EC keygen and ECDSA sign and verify support

williamcroberts:
* Probably want to copy the sign/verify bit to sign/verify tests as the sign/verify in keygen is really about ensuring that the objects function as intended post key generation.


williamcroberts:
OSSL pre 1.1 issues...

PR #196, opened 2019-04-10
object: make public objects a first class citizen

williamcroberts:
The linking mechanism made C_DestroyObject and certificate support
a nightmare.

Just make the public portion of an object a seperate and descrete object
in the db that uses only the public portion of a TPM object and can
be loaded to the null hierarchy via loadexternal.


williamcroberts:
@PeterHuewe this should make destroy much easier. I wish someone slapped me in the face across the internet with that linking scheme nightmare.

PR #217, opened 2019-05-09
Add ccache to speed up travis runs.

DomiNic5787:
see [https://github.com/tpm2-software/tpm2-tss/pull/1403](https://github.com/tpm2-software/tpm2-tss/pull/1403)

Signed-off-by: Dominic Grauvogl <dominicmanuel.grauvogl@infineon.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ services:
 compiler:
 - gcc
 - clang
+cache: ccache
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


diabonas:
Same comment as https://github.com/tpm2-software/tpm2-tools/pull/1470#issuecomment-490941567: I think there is more to set up in order for this to work with the Docker container used here for building.

PR #220, opened 2019-05-13
Use yaml.safe_load for importing yaml

### LINKED ISSUE #213: yaml.load without Loader=... is deprecated
Calling `yaml.load()` without the `Loader=` parameter is unsafe and deprecated since PyYAML 5.1. The affected files are `commandlets_keys.py`, `commandlets_store.py`, `commandlets_token.py` and `tpm2.py`.

From [PyYAML's wiki pages](https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation):

> Use of PyYAML's yaml.load function without specifying the Loader=... parameter, has been deprecated. In PyYAML version 5.1, you will get a warning, but the function will still work.

PeterHuewe:
According to the yaml documentation yaml.load is deprecated since PyYAML
5.1 without specifying the loader.
Since we are not doing special stuff, we can use the safe_load function
instead.

Fixes #213

### DIFF
--- a/tools/tpm2_pkcs11/commandlets_keys.py
+++ b/tools/tpm2_pkcs11/commandlets_keys.py
@@ -85,7 +85,7 @@ class NewKeyCommandBase(Command):
         # and populate the db with the data. This allows use of the public data
         # without needed to load any objects which requires a pin to do.
         #
-        y = yaml.load(tertiarypubdata)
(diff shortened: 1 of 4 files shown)


williamcroberts:
Working on build break

PeterHuewe:
#219 makes them working again :)

PR #223, opened 2019-05-14
Pad manufacturer id and library description and add nullpointer check

PeterHuewe:
According to the standard these fields must be padded with blanks and
should not be null terminated.
Also it makes sense to check whether the supplied pointer is not null
and add a test

### DIFF
--- a/src/lib/general.c
+++ b/src/lib/general.c
@@ -32,18 +32,21 @@
          }
 
 CK_RV general_get_info(CK_INFO *info) {
+    check_pointer(info);
 
-    static CK_INFO _info = {
+    CK_INFO _info = {
         .cryptokiVersion = CRYPTOKI_VERSION,
-        .manufacturerID = " "TPM2_TOKEN_MANUFACTURER,
+        .manufacturerID = TPM2_TOKEN_MANUFACTURER,
         .flags = 0,
-        .libraryDescription = " "LIBRARY_DESCRIPTION,
+        .libraryDescription = LIBRARY_DESCRIPTION,
         .libraryVersion = {
             /* TODO get from build VERSION */
             .major = 42,
             .minor = 42
         },
     };
+    str_pad(_info.manufacturerID, sizeof(_info.manufacturerID));
+    str_pad(_info.libraryDescription, sizeof(_info.libraryDescription));
 
     *info = _info;
 
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -32,6 +32,15 @@ static inline int min(size_t a, size_t b) {
     return a < b ? a : b;
 }
 
+/*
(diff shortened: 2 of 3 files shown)


src/lib/general.c
```diff
@@ -32,18 +32,21 @@
...
-        .libraryDescription = " "LIBRARY_DESCRIPTION,
+        .libraryDescription = LIBRARY_DESCRIPTION,
         .libraryVersion = {
             /* TODO get from build VERSION */
             .major = 42,
             .minor = 42
         },
     };
+    str_pad(_info.manufacturerID, sizeof(_info.manufacturerID));
+    str_pad(_info.libraryDescription, sizeof(_info.libraryDescription));
```
williamcroberts: If the field is static width can't we do this statically?

src/lib/utils.h
```diff
@@ -32,6 +32,15 @@ static inline int min(size_t a, size_t b) {
     return a < b ? a : b;
 }
 
+/*
+ * Pads the buffer 'buf' with blanks to the desired 'buf_len'.
+ * The trailing '\0' byte is removed, so the buffer is NOT null terminared afterwards!
+ * */
+static inline void str_pad(unsigned char * buf, size_t buf_len) {
+    size_t str_len = strlen((char *)buf);
+    memset(buf + str_len, ' ', buf_len - str_len);
```
williamcroberts: Wouldn't the existing string need to be moved to the end too?

src/lib/general.c
PeterHuewe: The problem is as always the 'fill with blanks and must not be null terminated' requirement.
So by assigning it directly, you either have it one  blank to less (and a null terminator), or you are basically storing the null terminator outside of the field.

An alternative is of course to specify each character individually, but this is hard on readability.

Using the method in the patch gets around the problem, as we can simply assign it (=readable) and then we pad it to the correct length and remove the null terminator.

src/lib/utils.h
PeterHuewe: all other pkcs11 strings we have are left aligned and padded to the right with blanks.
The standard does not really specify if left or right aligned.
We can  of course change it, but then we have to change str_padded_copy too.

PeterHuewe: https://lists.oasis-open.org/archives/pkcs11/201406/msg00010.html 
If I read this correctly, they must be left aligned and padded to the right

williamcroberts: before it was a static structure with no additional writes, thus making it thread safe, now it's not thread safe.

PeterHuewe: I see.
Remove the static to make it a local variable, which is thread safe again.?

williamcroberts: Yep

PeterHuewe: Done&updated

PR #226, opened 2019-05-16
Retrieve Token and Slot Info from TPM

PeterHuewe:
This patchset retrieves the token info from the TPM and maps the result also to the slot info.

Signed-off by: Peter Huewe <peter.huewe@infineon.com>

### DIFF
--- a/src/lib/slot.c
+++ b/src/lib/slot.c
@@ -73,26 +73,26 @@ CK_RV slot_get_list (CK_BYTE token_present, CK_SLOT_ID *slot_list, CK_ULONG_PTR 
 
 CK_RV slot_get_info (CK_SLOT_ID slot_id, CK_SLOT_INFO *info) {
 
-    const CK_BYTE manufacturerID[] = "foo";
-    const CK_BYTE slotDescription[] = "bar";
+    token *token;
+    CK_TOKEN_INFO token_info;
 
     check_pointer(info);
 
-    if (!slot_get_token(slot_id)) {
+    token = slot_get_token(slot_id);
+    if (!token) {
         return CKR_SLOT_ID_INVALID;
     }
 
     memset(info, 0, sizeof(*info));
 
-    /* TODO pull these from TPM */
-    info->firmwareVersion.major =
-            info->firmwareVersion.minor = 13;
+    if (token_get_info(token, &token_info)) {
+        return CKR_GENERAL_ERROR;
+    }
 
-    info->hardwareVersion.major =
-    info->hardwareVersion.minor = 42;
-
-    str_padded_copy(info->manufacturerID, manufacturerID, sizeof(info->manufacturerID));
-    str_padded_copy(info->slotDescription, slotDescription, sizeof(info->slotDescription));
+    str_padded_copy(info->manufacturerID, token_info.manufacturerID, sizeof(info->manufacturerID));
+    str_padded_copy(info->slotDescription, token_info.label, sizeof(info->slotDescription));
+    info->hardwareVersion = token_info.hardwareVersion;
+    info->firmwareVersion = token_info.firmwareVersion;
 
     info->flags = CKF_TOKEN_PRESENT | CKF_HW_SLOT;
 
--- a/src/lib/token.c
+++ b/src/lib/token.c
@@ -69,31 +69,18 @@ void token_free(token *t) {
 }
 
 CK_RV token_get_info (token *t, CK_TOKEN_INFO *info) {
-
+    check_pointer(t);
     check_pointer(info);
-
-    const CK_BYTE token_sn[]   = TPM2_TOKEN_SERIAL_NUMBER;
-    const CK_BYTE token_manuf[]  = TPM2_TOKEN_MANUFACTURER;
-    const CK_BYTE token_model[]  = TPM2_TOKEN_MODEL;
-    const CK_BYTE token_hwver[2] = TPM2_SLOT_HW_VERSION;
-    const CK_BYTE token_fwver[2] = TPM2_SLOT_FW_VERSION;
+    int rval;
     time_t rawtime;
     struct tm * tminfo;
 
     memset(info, 0, sizeof(*info));
 
-    /*
-     * TODO Set these to better values
-     * and get valid VERSION info. Likely
-     * need to make version match what is in general.c
-     * for the CK_INFO structure, not sure.
-     *
-     * Below is ALL the fields grouped.
-     */
-
-    // Version info
-    memcpy(&info->firmwareVersion, &token_fwver, sizeof(token_fwver));
-    memcpy(&info->hardwareVersion, &token_hwver, sizeof(token_hwver));
+    rval = tpm_get_token_info(t->tctx, info);
+    if (rval != CKR_OK) {
+        return CKR_GENERAL_ERROR;
+    }
 
     // Support Flags
     info->flags = CKF_RNG
@@ -105,9 +92,8 @@ CK_RV token_get_info (token *t, CK_TOKEN_INFO *info) {
 
     // Identification
     str_padded_copy(info->label, t->label, sizeof(info->label));
-    str_padded_copy(info->manufacturerID, token_manuf, sizeof(info->manufacturerID));
-    str_padded_copy(info->model, token_model, sizeof(info->model));
-    str_padded_copy(info->serialNumber, token_sn, sizeof(info->serialNumber));
+    str_padded_copy(info->serialNumber, (unsigned char*) TPM2_TOKEN_SERIAL_NUMBER, sizeof(info->serialNumber));
+
 
     // Memory: TODO not sure what memory values should go here, the platform?
     info->ulFreePrivateMemory = ~0;
--- a/src/lib/token.c
+++ b/src/lib/token.c
@@ -114,6 +114,9 @@ CK_RV token_get_info (token *t, CK_TOKEN_INFO *info) {
     time (&rawtime);
     tminfo = gmtime(&rawtime);
     strftime ((char *)info->utcTime, sizeof(info->utcTime), "%Y%m%d%H%M%S", tminfo);
+    // The last two bytes must be '0', not NULL/'\0' terminated.
+    info->utcTime[14] = '0';
+    info->utcTime[15] = '0';
 
     return CKR_OK;
 }
--- a/src/lib/tpm.c
+++ b/src/lib/tpm.c
@@ -11,6 +11,8 @@
 #include <string.h>
 #include <stddef.h>
 
+#include <arpa/inet.h>
+
 #include <openssl/asn1.h>
 #include <openssl/bn.h>
 #include <openssl/ec.h>
@@ -22,12 +24,26 @@
 #include <tss2/tss2_esys.h>
 #include <tss2/tss2_mu.h>
 
+#include "checks.h"
 #include "openssl_compat.h"
 #include "pkcs11.h"
 #include "log.h"
 #include "mutex.h"
 #include "tcti_ldr.h"
 #include "tpm.h"
+
+/**
+ * Maps 4-byte manufacturer identifier to manufacturer name.
+ */
+static const char *TPM2_MANUFACTURER_MAP[][2] = {
+    {"ATML", "Atmel"},
+    {"INTC", "Intel"},
+    {"IFX ", "Infineon"},
+    {"IBM ", "IBM"},
+    {"NTC ", "Nuvoton"},
+    {"STM ", "STMicro"}
+};
+
 
 struct tpm_ctx {
     TSS2_TCTI_CONTEXT *tcti_ctx;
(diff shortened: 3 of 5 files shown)


src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
...
+    }
+
+    if (!capabilityData || capabilityData->data.tpmProperties.count < TPM2_PT_VENDOR_STRING_4 - TPM2_PT_FIXED + 1) {
+        LOGE("TPM did not reply with correct amount of capabilities");
+	rv = CKR_GENERAL_ERROR;
+	goto out;
+    }
+
+    // Use Spec revision as HW Version
+    revision = capabilityData->data.tpmProperties.tpmProperty[TPM2_PT_REVISION - TPM2_PT_FIXED].value;
```
williamcroberts: revision should be UINT32 also declare at smallest scope, here and throughout.

src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
...
+
+    // Use Vendor String as Model
+    manufacturer = ntohl(capabilityData->data.tpmProperties.tpmProperty[TPM2_PT_MANUFACTURER - TPM2_PT_FIXED].value);
+    memcpy(manufacturerID, (unsigned char*) &manufacturer, sizeof(uint32_t));
+    str_padded_copy(info->manufacturerID, manufacturerID, sizeof(info->manufacturerID));
+
+    // Map human readable Manufacturer String, if available
+    for (unsigned int i=0; i < ARRAY_LEN(TPM2_MANUFACTURER_MAP); i++){
+        if (strncasecmp((char *)info->manufacturerID, (char *)TPM2_MANUFACTURER_MAP[i][0], 4) == 0){
+            str_padded_copy(info->manufacturerID, (unsigned char *)TPM2_MANUFACTURER_MAP[i][1], sizeof(info->manufacturerID));
```
williamcroberts: if we do this, we need a default of just putting the bytes in there.

src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
...
+        ESYS_TR_NONE,
+        capability,
+        property, propertyCount, &moreData, &capabilityData);
+
+    if (rval != TSS2_RC_SUCCESS) {
+        LOGE("Esys_GetCapability: 0x%x:", rval);
+        return CKR_GENERAL_ERROR;
+    }
+
+    if (!capabilityData || capabilityData->data.tpmProperties.count < TPM2_PT_VENDOR_STRING_4 - TPM2_PT_FIXED + 1) {
```
williamcroberts: some of these seem overlength....

src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
...
+    uint32_t version;
+    uint32_t manufacturer;
+    unsigned char manufacturerID[5] = {0}; // 4 bytes + '\0'
+    uint32_t vendor[4];
+    CK_RV rv = CKR_OK;
+    TPM2_CAP capability = TPM2_CAP_TPM_PROPERTIES;
+    UINT32 property = TPM2_PT_FIXED;
+    UINT32 propertyCount = TPM2_MAX_TPM_PROPERTIES;
+    TPMS_CAPABILITY_DATA *capabilityData;
+    TPMI_YES_NO moreData;
```
williamcroberts: declare closest to use. Also, uint32_t should be UINT32. Always match to the TPM type where appropriate.

src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
 LOAD_TYPE(TPM2B_PUBLIC, public)
 LOAD_TYPE(TPM2B_PRIVATE, private)
 
+CK_RV tpm_get_token_info_from_tpm (tpm_ctx *ctx, CK_TOKEN_INFO *info) {
+
+    uint32_t revision;
+    uint32_t version;
+    uint32_t manufacturer;
+    unsigned char manufacturerID[5] = {0}; // 4 bytes + '\0'
+    uint32_t vendor[4];
```
williamcroberts: Anyway to remove the magic buffer size here?

src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
...
+
+    // Use Firmware Version as FW Version
+    version = capabilityData->data.tpmProperties.tpmProperty[TPM2_PT_FIRMWARE_VERSION_1 - TPM2_PT_FIXED].value;
+    // Most vendors seem to use 00MM.00mm as format for TPM2_PT_FIRMWARE_VERSION_1
+    // Unfortunately we only have 1 byte for major and minor each.
+    info->firmwareVersion.major = (version >> 16) & 0xFF;
+    info->firmwareVersion.minor = version  & 0xFF;
+
+    // Use Vendor String as Model
+    manufacturer = ntohl(capabilityData->data.tpmProperties.tpmProperty[TPM2_PT_MANUFACTURER - TPM2_PT_FIXED].value);
```
williamcroberts: so size on value should be what your expecting right? At least an assert() here would make sense. What if the data  coming back is over length ie TPM MITM?

src/lib/token.c
```diff
@@ -128,6 +114,9 @@ CK_RV token_get_info (token *t, CK_TOKEN_INFO *info) {
     time (&rawtime);
     tminfo = gmtime(&rawtime);
     strftime ((char *)info->utcTime, sizeof(info->utcTime), "%Y%m%d%H%M%S", tminfo);
+    // The last two bytes must be '0', not NULL/'\0' terminated.
```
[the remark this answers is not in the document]
PeterHuewe: No. pkcs11 strings are usually never null terminated :)
`info->utcTime` is specified as `CK_CHAR utcTime[16];`

In the spec under http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc323024051 `utcTime `is specified as :
> current time as a character-string of length 16, represented in the format YYYYMMDDhhmmssxx (4 characters for the year;  2 characters each for the month, the day, the hour, the minute, and the second; and 2 additional reserved ‘0’ characters).  The value of this field only makes sense for tokens equipped with a clock, as indicated in the token information flags 

So we have YYYY MMDD hhmm ssxx = 16 characters, where the last two are `2 additional reserved ‘0’ characters`, where there did want to have some space for subsecond time resolution.

No space for NULL termination (as with every other pkcs11 string ;)

src/lib/tpm.c
```diff
@@ -22,13 +24,24 @@
...
 #include "pkcs11.h"
 #include "log.h"
 #include "mutex.h"
 #include "tcti_ldr.h"
 #include "tpm.h"
 
+/**
+ * Maps 4-byte manufacturer identifier to manufacturer name.
+ */
+static const char *TPM2_MANUFACTURER_MAP[][2] = {
```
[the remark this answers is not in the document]
PeterHuewe: I thought also about this while writing.

Since PKCS11 is a 'high-level' interface, I guess users would expect something high level as vendor strings, not cryptic 4 bytes abreviations. (who is the TPM Vendor 'WEC' for example, INTC is also not too obvious.)
Since Infineon is also publicly known as IFX, I don't really have a problem in keeping the reported short handle (of IFX in my case).

We could incoporate the list of all (18) TPM vendors from the TCG Vendor Registry, https://trustedcomputinggroup.org/wp-content/uploads/Vendor_ID_Registry_0-8_clean.pdf 
which has been last updated 2015, so it does not seem to be that volatile and thus not too hard to maintain.

Since some of them are not really manufacturing any TPM2.0 chips, I would rather have the list grow, when they are really working and tested with our provider.

Maybe see being in the list as a benefit for active contributors.
I can also split out this section into a seperate patch where we can extend the discussion, as I really want to see the general thing merged.

src/lib/slot.c
```diff
@@ -73,26 +73,26 @@ CK_RV slot_get_list (CK_BYTE token_present, CK_SLOT_ID *slot_list, CK_ULONG_PTR
...
 
-    if (!slot_get_token(slot_id)) {
+    token = slot_get_token(slot_id);
+    if (!token) {
         return CKR_SLOT_ID_INVALID;
     }
 
     memset(info, 0, sizeof(*info));
 
-    /* TODO pull these from TPM */
```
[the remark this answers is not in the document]
PeterHuewe: I think the current approach is sane, especially as you state 'the TPM is the token' (which is plugged into a slot).
So if you ask the slot about its information, it looks into its contents (i.e. the token) and retrieved the info from there.

src/lib/tpm.c
```diff
@@ -486,6 +499,77 @@ bool files_load_bytes_from_path(const char *path, UINT8 *buf, UINT16 *size) {
 LOAD_TYPE(TPM2B_PUBLIC, public)
 LOAD_TYPE(TPM2B_PRIVATE, private)
 
+CK_RV tpm_get_token_info_from_tpm (tpm_ctx *ctx, CK_TOKEN_INFO *info) {
```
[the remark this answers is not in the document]
PeterHuewe: ok - I was not too sure about the naming convention.
I would suggest tpm_get_token_info though (as it retrieves the information into the token_info struct).

src/lib/tpm.c
PeterHuewe: ok, no problem

src/lib/tpm.c
PeterHuewe: Which is exactly done in the memcpy + str_padded_copy in the lines above. (551f)
So only if there is a match in the table, the bytes will be modified, otherwise we just have the short hand identifier padded with blanks.

src/lib/tpm.c
PeterHuewe: sizeof(UINT32)+1

src/lib/tpm.c
PeterHuewe: ok

src/lib/tpm.c
PeterHuewe: in terms of line lenght - yes.
I'll split them into two lines if it improves readability.
Coming from kernel space, I know how code sometimes looks if the line length is strictly enforced...
Maybe it would make sense to document somewhere what our coding style should look like (over all tpm2-software projects).

src/lib/tpm.c
PeterHuewe: value is defined in the struct as an UINT32, so 4 bytes. 
These 4 bytes do NOT contain any length information, they are always 4 bytes.
If the data coming from the TPM is tampered in any way, we would read here exactly 4 byte (of garbage), but it would not harm in any way.

PeterHuewe:
@williamcroberts  I uploaded a new version which should address your requested changes.
I left the TCG Vendor map in, if you really don't like it I can respin one more time.

src/lib/token.c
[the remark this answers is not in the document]
williamcroberts: Sorry read that as byte value 0 not ASCII 0.

src/lib/tpm.c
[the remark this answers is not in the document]
williamcroberts: Im ok with the map, we can try it. Just make sure we have a default that takes the raw values and shoves them in there.

src/lib/tpm.c
```diff
@@ -22,13 +24,24 @@
...
 #include "tcti_ldr.h"
 #include "tpm.h"
 
+/**
+ * Maps 4-byte manufacturer identifier to manufacturer name.
+ */
+static const char *TPM2_MANUFACTURER_MAP[][2] = {
+    {"INTC", "Intel"},
+    {"IFX ", "Infineon"},
+    {"IBM ", "IBM"},
```
williamcroberts: expand the map to the best of your ability to the registry and ensure a default handler to just plop their data in there.

williamcroberts:
Overall looks good, I tried it out locally, i have one question: can we use serial for something?
```
Token 1:
	URL: pkcs11:model=SW%20%20%20TPM;manufacturer=IBM;serial=0000000000000000;token=label
	Label: label
	Type: Hardware token
	Manufacturer: IBM
	Model: SW   TPM
	Serial: 0000000000000000
	Module: libtpm2_pkcs11.so
```

src/lib/tpm.c
PeterHuewe: The default *is* handled in the current code.
By expanding you mean adding the other entries from tcg?

src/lib/slot.c
```diff
@@ -73,26 +73,26 @@ CK_RV slot_get_list (CK_BYTE token_present, CK_SLOT_ID *slot_list, CK_ULONG_PTR
...
-    info->firmwareVersion.major =
-            info->firmwareVersion.minor = 13;
-
-    info->hardwareVersion.major =
-    info->hardwareVersion.minor = 42;
+    if (token_get_info(token, &token_info) != CKR_OK) {
```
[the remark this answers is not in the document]
PeterHuewe: Ok.
I was not aware there is any coding standard.
The rationale was to use a variable if the value is used later on (l81 change)

and direct call if the (temp) variable would not be used later on.

I'll change it never the less

PeterHuewe:
> Overall looks good (minor things), I tried it out locally, i have one question: can we use serial for something?
Not sure. The TPM has specifically no serial number due to privacy.

src/lib/tpm.c
PeterHuewe: I updated it to the tpm2.0 vendors known to me.

src/lib/slot.c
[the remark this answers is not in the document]
PeterHuewe: Ah, I see, implicit vs explicit checking.
Fixed

PeterHuewe:
@williamcroberts  New version available

williamcroberts:
Fix the test

PeterHuewe:
@williamcroberts  Done - sorry for failing the tests.
By changing from explicit to implicit check, I inverted the logic :/
This is why I'm not a too big fan of the implicit way of checking stuff.

PeterHuewe:
Merge, as discussed via email.

PR #227, opened 2019-05-16
mechanism: Report only mechanisms supported by TPM

PeterHuewe:
Query the TPM whether a specified mechanism can be supported or not.
Unfortunately the mapping is not 1:1.
There might be cases where the combination of supported algorithms by
the TPM show that a mechanism is supported, but eventually cannot be
used for *certain* use cases.
(e.g. `aes` reported as supported, but `TPM2_EncryptDecrypt` is not
implemented, then we cannot use it for encrypt or decrypt).
These cases can maybe be specified/handled in more detail in
`slot_mechanism_info_get`.

This logic atleast filters out mechanism, which are definitely not
supported by the specific TPM (e.g. `sha512`).
 
### DIFF
--- a/src/lib/slot.c
+++ b/src/lib/slot.c
@@ -99,50 +99,15 @@ CK_RV slot_get_info (CK_SLOT_ID slot_id, CK_SLOT_INFO *info) {
     return CKR_OK;
 }
 
-static const CK_MECHANISM_TYPE mechs[] = {
-    CKM_AES_CBC,
-    CKM_AES_CFB1,
-    CKM_AES_ECB,
-    CKM_ECDSA,
-    CKM_ECDSA_SHA1,
-    CKM_EC_KEY_PAIR_GEN,
-    CKM_RSA_PKCS,
-    CKM_RSA_PKCS_KEY_PAIR_GEN,
-    CKM_RSA_PKCS_OAEP,
-    CKM_RSA_X_509,
-    CKM_SHA_1,
-    CKM_SHA1_RSA_PKCS,
-    CKM_SHA256,
-    CKM_SHA256_RSA_PKCS,
-    CKM_SHA384,
-    CKM_SHA384_RSA_PKCS,
-    CKM_SHA512,
-    CKM_SHA512_RSA_PKCS,
-};
 
 CK_RV slot_mechanism_list_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE *mechanism_list, CK_ULONG_PTR count) {
-
-    if (!slot_get_token(slot_id)) {
+    token *t = slot_get_token(slot_id);
+    if (!t) {
         return CKR_SLOT_ID_INVALID;
     }
 
-    if (!count){
-        return CKR_ARGUMENTS_BAD;
-    }
-
-    if (!mechanism_list) {
-        *count = ARRAY_LEN(mechs);
-        return CKR_OK;
-    }
-
-    if (*count < ARRAY_LEN(mechs)) {
-        return CKR_BUFFER_TOO_SMALL;
-    }
-
-    *count = ARRAY_LEN(mechs);
-    memcpy(mechanism_list, mechs, sizeof(mechs));
-
-    return CKR_OK;
+    CK_RV rv = tpm2_getmechanisms(t->tctx, mechanism_list, count);
+    return rv;
 }
 
 CK_RV slot_mechanism_info_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE type, CK_MECHANISM_INFO *info) {
--- a/src/lib/tpm.h
+++ b/src/lib/tpm.h
@@ -178,4 +178,6 @@ CK_RV tpm2_generate_key(
 
         tpm_object_data *objdata);
 
+CK_RV tpm2_getmechanisms(tpm_ctx *ctx, CK_MECHANISM_TYPE *mechanism_list, CK_ULONG_PTR count);
+
 #endif /* SRC_PKCS11_TPM_H_ */
--- a/src/lib/tpm.c
+++ b/src/lib/tpm.c
@@ -2780,3 +2780,148 @@ void tpm_objdata_free(tpm_object_data *objdata) {
     }
 
 }
+
+static const CK_MECHANISM_TYPE mechs[] = {
+    CKM_AES_CBC,
+    CKM_AES_CFB1,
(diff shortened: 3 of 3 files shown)


PeterHuewe:
Not sure why travis failed.
I was testing it locally with a newer version of the simulator (1134) - I will re-try with 1119

src/lib/slot.c
```diff
@@ -139,8 +144,134 @@ CK_RV slot_mechanism_list_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE *mechanism_
...
+            case CKM_SHA384_RSA_PKCS:
+                if (is_algorithm_supported(algs, TPM2_ALG_SHA384) &&
+                    is_algorithm_supported(algs, TPM2_ALG_RSA)) {
+                    mechanism_list[supported] = mechs[i];
+                    supported++;
+                }
+                break;
+            case CKM_SHA512:
+                if (is_algorithm_supported(algs, TPM2_ALG_SHA512)) {
+                    mechanism_list[supported] = mechs[i];
```
PeterHuewe: Interestingly, code coverage shows that the tpm simulator does not support sha512 - so sha512 is 'thoroughly tested' :P

PeterHuewe:
I also planned to remove the `is_mech_supported` logic in `sign.c`, but this breaks the tests for `CKM_SHA512_RSA_PKCS` in `pkcs-sign-verify.int.c` which is emulated in **software**.
Of course the comment above `is_mech_supported ``XXX this is probably best to query the TPM layer` is wrong then.
Not sure how to deal with this, whether we want to keep the sw hash or not.
Keeping it, makes the querying from TPM for the slot mechanism harder, as we have to keep track which algorithms are sw emulated - but offers advanced functionality.

You can find the commit here https://github.com/PeterHuewe/tpm2-pkcs11/commit/ec348b65cbf8e946306cffcddd99df4698607d43 if interested.

williamcroberts:
I left a comment on your patch https://github.com/PeterHuewe/tpm2-pkcs11/commit/ec348b65cbf8e946306cffcddd99df4698607d43

src/lib/slot.c
```diff
@@ -121,16 +122,20 @@ static const CK_MECHANISM_TYPE mechs[] = {
 };
 
 CK_RV slot_mechanism_list_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE *mechanism_list, CK_ULONG_PTR count) {
+    int supported = 0;
```
williamcroberts: CK_ULONG

src/lib/slot.c
```diff
@@ -139,8 +144,23 @@ CK_RV slot_mechanism_list_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE *mechanism_
         return CKR_BUFFER_TOO_SMALL;
     }
 
-    *count = ARRAY_LEN(mechs);
-    memcpy(mechanism_list, mechs, sizeof(mechs));
+    TPMS_CAPABILITY_DATA *capabilityData = NULL;
+    rv = tpm_get_algorithms (t->tctx, &capabilityData);
```
williamcroberts: I wonder if this would be better on C_Init() and then cache it.

src/lib/slot.c
```diff
@@ -139,8 +144,23 @@ CK_RV slot_mechanism_list_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE *mechanism_
...
+    if (rv != CKR_OK) {
+        LOGE("Retrieving supported algorithms from TPM failed");
+        return rv;
+    }
+    TPMU_CAPABILITIES *algs= &capabilityData->data;
+
+    for (unsigned int i = 0; i < ARRAY_LEN(mechs); i++) {
+        if (tpm_algs_is_mechanism_supported(algs, mechs[i])) {
```
williamcroberts: Im confused by this interface, hey TPM what algorithms you got, are they supported? I would imagine the interface to the tpm layer should just be, give me a list of pkcs11 mechs you support. Rather than this we have a list, now filter.

src/lib/tpm.h
```diff
@@ -163,4 +165,8 @@ CK_RV tpm2_generate_key(
 
         tpm_object_data *objdata);
 
+CK_RV tpm_get_algorithms (tpm_ctx *ctx, TPMS_CAPABILITY_DATA **capabilityData);
+CK_BBOOL is_algorithm_supported(TPMU_CAPABILITIES *capabilities, TPM2_ALG_ID algorithm);
```
williamcroberts: We leave the non-tpm layer tpm type free

src/lib/tpm.h
```diff
@@ -163,4 +165,8 @@ CK_RV tpm2_generate_key(
...
+CK_BBOOL tpm_algs_is_mechanism_supported(TPMU_CAPABILITIES *algs, CK_MECHANISM_TYPE mechanism);
```
williamcroberts: ETOOCOMPLICATED....

just something like:
```C
tpm2_getmechanisms(tpm_ctx *ctx, CK_MECHANISM_TYPE *mechanism_list, CK_ULONG_PTR count);
```
Call it on initialization of the library, cache the result and return it.

src/lib/slot.c
PeterHuewe: The list is what the library supports in general, which is then filtered by the TPM capabilities.
That's the rationale behind this idea.
We can hide it, okay.

src/lib/slot.c
PeterHuewe: The general question is, what shall we cache? (in the context?)
The algorithms capability?, the supported mechanism?, the mechanism+details?

williamcroberts: The overall current approach is awkward even if you don't just cache the data structures on init and do it on demand. None of these structures then would be a static list you filter non-supported things out of. Just always ask the TPM and return the values. If someone doesn't provide size, ask the TPM and return the list count. Add caching for bonus points.

PeterHuewe:
@williamcroberts Please have a look at this and #240  
Personally I would prefer this version, but either one is fine.

williamcroberts:
No #240 is much better. Just look at the LOC delta.

#227 +151 -39 3 files changed
#240 +123 -39 3 files changed

I expect 240 to grow, but the song and dance between the pkcs11 layer and the tpm is just avoided. You just ask the hardware what is supported rather than generating and filtering.

PeterHuewe:
Works for me :)

PR #231, opened 2019-05-21
build: remove tss2-tcti-{device,mssim} dependencies

diabonas:
According to `ldd -u -r` these libraries are not used and can be removed. Furthermore `TCTI_SOCKET_LIBS` isn't defined anywhere, so remove it as well.


configure.ac
```diff
@@ -59,8 +59,6 @@ AC_CONFIG_HEADERS([src/lib/config.h])
 # test for TSS dependecies
 PKG_CHECK_MODULES([TSS2_ESYS],   [tss2-esys])
 PKG_CHECK_MODULES([TSS2_MU],     [tss2-mu])
-PKG_CHECK_MODULES([TCTI_DEVICE], [tss2-tcti-device])
-PKG_CHECK_MODULES([TCTI_MSSIM],  [tss2-tcti-mssim])
```
williamcroberts: Yeah these are dlopen() so they don't have direct dependencies.

williamcroberts:
Please deal with conflicts

diabonas:
Done!

Issue #201, opened 2019-04-14
Calling C_Sign for size doesn't work reliably for ECDSA


PeterHuewe:
Although the basic functionality was implemented/fixed in #133, as per discussion in
https://github.com/PeterHuewe/tpm2-pkcs11/commit/7f5f858581105a17b71f77cdd251b1137032563a
there are cases, where the returned size does not match the length of the signature.

So by calling C_Sign for size might return something like 34+34+2 = 71 bytes,
whereas the real signature is then 35+34+2 bytes.

The ASN1 Encoding of a ECDSA Signature is:
```
SEQUENCE (2 elem)
  INTEGER R 
  INTEGER S
```

If either INTEGER R or S have the highest bit set, an additional `00` byte is prepended in order to ensure a positive number.
(See https://docs.microsoft.com/de-de/windows/desktop/SecCertEnroll/about-integer for explanation).

So the size might change between two calls, which causes the signature operation to fail.

Possible solutions:
* Store the signature already in the first call for size and just return it once a suitable buffer is allocated.
* just add two bytes to the returned required buffer size
* add the leading 00 consistently to the integers

The latter also might be helpful for some applications which expect to have R and S of equal size.

williamcroberts:
just take ((keysize in bytes + hdr + extra) * 2) + seqhdr;

((32 + 2 + 1) * 2) + 2 = 72

Wouldn't that work? Then we don't have to go through the headache of caching and computing off of cache?

PeterHuewe:
Yes - that's the second option.
Would work for me.

PeterHuewe:
Unfortunately the fix does not work in all cases, i.e. if `opdata->do_hash` is set to false i.e. for 'raw' ECDSA (without SHA1)

PR #239, opened 2019-05-28
Fix Buffer Size for ECDSA

PeterHuewe:
The function ec_fixup_size was not implemented correctly,
as the initial logic whether the fixup should be applied was inverted.
Moreover the function was only called if do_hash was set to true,
which is not the case for 'raw' ECDSA.

Fixes: 62a521d4964c8af03d886920be52047b4d60fe44
Fixes: #201

### DIFF
--- a/src/lib/sign.c
+++ b/src/lib/sign.c
@@ -96,8 +96,8 @@ static bool is_mech_supported(CK_MECHANISM_TYPE mech) {
 
 static CK_RV ec_fixup_size(CK_MECHANISM_TYPE mech, tobject *tobj, CK_ULONG_PTR signature_len) {
 
-    if (mech != CKM_ECDSA
-            || mech != CKM_ECDSA_SHA1) {
+    if (mech != CKM_ECDSA &&
+        mech != CKM_ECDSA_SHA1) {
         /* nothing to fix up */
         return CKR_OK;
     }
@@ -537,6 +537,14 @@ CK_RV sign_final_ex(session_ctx *ctx, CK_BYTE_PTR signature, CK_ULONG_PTR signat
      */
     reset_ctx = (rv == CKR_BUFFER_TOO_SMALL || !signature);
     if (reset_ctx) {
+        /* ec signature size is not stable between calls, fix it up */
+        CK_RV tmp; /* we must not overwrite rv */
+        tmp = ec_fixup_size(opdata->mtype, opdata->tobj, signature_len);
+        if (tmp != CKR_OK) {
+            reset_ctx = false;
+            goto session_out;
+        }
+
         if (opdata->do_hash) {
             /* reset the hashing state */
             digest_op_data *new_digest_state = digest_op_data_new();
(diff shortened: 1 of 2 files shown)


PeterHuewe:
I'll look into why the tests fail....

williamcroberts:
> I'll look into why the tests fail....

Probably add a regression test too. We need to get in the habit of adding a test when we fix a bug.

PeterHuewe:
Same problem as always.
if (rv) vs if(!ptr) ....

williamcroberts:
> Same problem as always.
> if (rv) vs if(!ptr) ....

I've done that myself, many times.

PeterHuewe:
> 
> 
> > Same problem as always.
> > if (rv) vs if(!ptr) ....
>> if (rv) vs if(!ptr) ....
> 
> I've done that myself, many times.

That's why I prefer the explicit version for regular return codes.

williamcroberts:
> > > Same problem as always.
> > > if (rv) vs if(!ptr) ....
> > > if (rv) vs if(!ptr) ....
> > 
> > 
> > I've done that myself, many times.
> 
> That's why I prefer the explicit version for regular return codes.

Why'd you change it from:
https://github.com/tpm2-software/tpm2-pkcs11/pull/239/files#diff-c3df73929e97a82485422329ed256f2cL563
```C
            tmp = ec_fixup_size(opdata->mtype, opdata->tobj, signature_len);
            if (tmp != CKR_OK) {
```

PeterHuewe:
> 
> 
> > > > Same problem as always.
> > > > if (rv) vs if(!ptr) ....
> > > > if (rv) vs if(!ptr) ....
> > > 
> > > 
> > > I've done that myself, many times.
> > 
> > 
> > That's why I prefer the explicit version for regular return codes.
> 
> Why'd you change it from:
> https://github.com/tpm2-software/tpm2-pkcs11/pull/239/files#diff-c3df73929e97a82485422329ed256f2cL563
> 
> ```c
>             tmp = ec_fixup_size(opdata->mtype, opdata->tobj, signature_len);
>             if (tmp != CKR_OK) {
> ```

To comply with the "correct coding style/standard" according to you:
https://github.com/tpm2-software/tpm2-pkcs11/pull/226#discussion_r285188439

williamcroberts:
> > > > > Same problem as always.
> > > > > if (rv) vs if(!ptr) ....
> > > > > if (rv) vs if(!ptr) ....
> > > > 
> > > > 
> > > > I've done that myself, many times.
> > > 
> > > 
> > > That's why I prefer the explicit version for regular return codes.
> > 
> > 
> > Why'd you change it from:
> > https://github.com/tpm2-software/tpm2-pkcs11/pull/239/files#diff-c3df73929e97a82485422329ed256f2cL563
> > ```c
> >             tmp = ec_fixup_size(opdata->mtype, opdata->tobj, signature_len);
> >             if (tmp != CKR_OK) {
> > ```
> 
> To comply with the "correct coding style/standard" according to you:
> [#226 (comment)](https://github.com/tpm2-software/tpm2-pkcs11/pull/226#discussion_r285188439)

I meant to not bury the function call in a conditional statement. It has nothing to do with the the conditional check. if(x) and if (x != OK) are both OK however if (foo()) or if (foo != OK) are NOT OK.

PeterHuewe:
> > I'll look into why the tests fail....
> 
> Probably add a regression test too. We need to get in the habit of adding a test when we fix a bug.
Done - I modified the existing test cases that they test that the reported size is larger or equal the actual size.

PR #240, opened 2019-05-29
mechanism: Report only mechanisms supported by TPM (alternative)

PeterHuewe:
[This is an alternative version which does not filter the list, but might be harder to maintain]

Query the TPM whether a specified mechanism can be supported or not.
Unfortunately the mapping is not 1:1.
There might be cases where the combination of supported algorithms by
the TPM show that a mechanism is supported, but eventually cannot be
used for *certain* use cases.
(e.g. `aes` reported as supported, but `TPM2_EncryptDecrypt` is not
implemented, then we cannot use it for encrypt or decrypt).
These cases can maybe be specified/handled in more detail in
`slot_mechanism_info_get`.

This logic atleast filters out mechanism, which are definitely not
supported by the specific TPM (e.g. `sha512`).

### DIFF
--- a/src/lib/slot.c
+++ b/src/lib/slot.c
@@ -99,50 +99,15 @@ CK_RV slot_get_info (CK_SLOT_ID slot_id, CK_SLOT_INFO *info) {
     return CKR_OK;
 }
 
-static const CK_MECHANISM_TYPE mechs[] = {
-    CKM_AES_CBC,
-    CKM_AES_CFB1,
-    CKM_AES_ECB,
-    CKM_ECDSA,
-    CKM_ECDSA_SHA1,
-    CKM_EC_KEY_PAIR_GEN,
-    CKM_RSA_PKCS,
-    CKM_RSA_PKCS_KEY_PAIR_GEN,
-    CKM_RSA_PKCS_OAEP,
-    CKM_RSA_X_509,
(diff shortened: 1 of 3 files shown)


williamcroberts:
As for
> (e.g. aes reported as supported, but TPM2_EncryptDecrypt is not
implemented, then we cannot use it for encrypt or decrypt).

You could check the config value in the db or just issue another getcap to see if the command interface is present before reporting it.

Perhaps it would be better to populate this data in the db from the python script. Then we pick up the caching. When we implement C_Initialize, we'll have to replicate it in C, but that's no big deal.

PeterHuewe:
Thanks for the comment.
I would propose to do the changes in a second PR.
The current PR vastly improves the situation already.

williamcroberts:
> Thanks for the comment.
> I would propose to do the changes in a second PR.
> The current PR vastly improves the situation already.

Fair enough, file a bug for that and plop the number in this thread and then merge at your leisure.

PeterHuewe:
Done - see #245 
Also by looking at it in more detail - the mechanism itself SHOULD be reported as supported, but only certain use cases should not be reported as supported. 
-> This should be handled in the the mechanism info function, which specifies the details of the usage.
There also it can be specified whether something is HW vs software.

PR #241, opened 2019-05-29
Remove (dead) file handling code from tpm.c

PeterHuewe:
The file handling functions in tpm.c are not used anywhere and are untouched
since the initial commit.
If they are needed in the future we can add them back (maybe in their own file).


williamcroberts:
I heart negative diffstats

PR #244, opened 2019-05-30
test: Fix true/false assignment

PeterHuewe:
A variable with the name ck_false should have the value CK_FALSE,
not CK_TRUE!

### DIFF
--- a/test/integration/pkcs-keygen.int.c
+++ b/test/integration/pkcs-keygen.int.c
@@ -153,7 +153,7 @@ static void test_rsa_keygen_missing_attributes(void **state) {
     verify_missing_pub_attrs_common(session, CKK_RSA, pub_handle_dup);
     verify_missing_pub_attrs_rsa(session, pub_handle_dup);
 
-    verify_missing_priv_attrs_common(session, CKK_RSA, priv_handle_dup);
+    verify_missing_priv_attrs_common(session, CKK_RSA, priv_handle_dup, CK_FALSE);
     verify_missing_priv_attrs_rsa(session, priv_handle_dup);
 }
 static void test_rsa_keygen_p11tool_templ(void **state) {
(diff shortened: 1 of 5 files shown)


test/integration/pkcs-keygen.int.c
```diff
@@ -63,7 +63,7 @@ static void test_rsa_keygen_p11tool_templ(void **state) {
     CK_SESSION_HANDLE session = ti->handle;
 
     CK_BBOOL ck_true = CK_TRUE;
-    CK_BBOOL ck_false = CK_TRUE;
+    CK_BBOOL ck_false = CK_FALSE;
```
williamcroberts: maybe ;-)

test/integration/pkcs-keygen.int.c
```diff
@@ -298,7 +298,7 @@ static void test_destroy(void **state) {
     CK_SESSION_HANDLE session = ti->handle;
 
     CK_BBOOL ck_true = CK_TRUE;
-    CK_BBOOL ck_false = CK_TRUE;
+    CK_BBOOL ck_false = ck_false;
```
williamcroberts: Looks like this is used for CKA_SENSITIVE, weird that it's causing it to fail....

PeterHuewe: not really weird.
The template sets CKA_SENSITIVE to CK_FALSE, which then sets CKA_ALWAYS_SENSITIVE also to false.
verify_missing_priv_attrs_common in test.c however expects CKA_ALWAYS_SENSITIVE to be true.

PeterHuewe: @williamcroberts  However I'm not sure what is the correct way to fix this.
Change the template to set it to CK_TRUE, (which is not equal to the p11 template?) or change verify_missing_priv_attrs_common (to check the existence of the attribute, but not the value itself, probably needed for CKA_EXTRACTABLE as well)

williamcroberts: > @williamcroberts However I'm not sure what is the correct way to fix this.
> Change the template to set it to CK_TRUE, (which is not equal to the p11 template?) or change verify_missing_priv_attrs_common (to check the existence of the attribute, but not the value itself, probably needed for CKA_EXTRACTABLE as well)

I want it to match the p11 template, so I would adjust the test.

PeterHuewe: And of course it uncovered a bug #246 :)

Issue #198, opened 2019-04-10
mechanism: support proper mechanism querying


williamcroberts:
- [x] slot_mechanism_info_get: doesn't actually return correct information to what the tpm can support.
- [ ] slot_mechanism_list_get: doesn't actually query the TPM for supported mechanisms.

PeterHuewe:
I think we already covered the first part :)

williamcroberts:
Seems fixed in: f5275084896ba

PR #248, opened 2019-06-03
Populate MECHANISM_INFO for supported mechanisms

PeterHuewe:
This patch populates the MECHANISM_INFO structure for each of the supported
mechanisms, as a preparation to pull this information from the TPM.

The min/max keysizes were taken from the `Implementation.h` from the simulator,
for now.
Maybe RSA1k keys should be removed from the list, as e.g. TCG PTP spec lists
them as 'support for 1024-bits keys is not recommended'.
The AES Keysizes are in bytes according to the standard, not bits, so this also
fixes a bug.

The usage flags were largely derived from tables like
http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850404

### DIFF
--- a/src/lib/slot.c
+++ b/src/lib/slot.c
@@ -118,23 +118,78 @@ CK_RV slot_mechanism_info_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE type, CK_ME
         return CKR_SLOT_ID_INVALID;
     }
 
-    // TODO support more of these and check with the TPM for sizes.
+    /* TODO pull these from TPM, currently they match the simulator */
+    CK_ULONG aes_min_keysize = 128/8; // in bytes
+    CK_ULONG aes_max_keysize = 256/8; // in bytes
+    CK_ULONG ecc_min_keysize = 256;
+    CK_ULONG ecc_max_keysize = 384;
+    CK_ULONG rsa_min_keysize = 1024;
+    CK_ULONG rsa_max_keysize = 2048;
+
     switch(type) {
+    /* AES based crypto */
+    /* Todo: Check if HW or Software and support */
     case CKM_AES_KEY_GEN:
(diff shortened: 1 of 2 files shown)


PeterHuewe:
Related to #198

PeterHuewe:
The codecov report was unfortunately not updated, code coverage does not decrease.

src/lib/slot.c
```diff
@@ -118,23 +118,78 @@ CK_RV slot_mechanism_info_get (CK_SLOT_ID slot_id, CK_MECHANISM_TYPE type, CK_ME
         return CKR_SLOT_ID_INVALID;
     }
 
-    // TODO support more of these and check with the TPM for sizes.
+    /* TODO pull these from TPM, currently they match the simulator */
```
williamcroberts: Let's not do this halfway, lets go to the TPM and run some getcap commands and actually look at what is supported. Not everything actually supports AES.

PeterHuewe: I would suggest one step at a time. I want to split this up in reviewable chunks.
This definitely improves the situation, as the fields are atleast populated, with sane values.
The real fix comes then when #198 gets closed.

For most of these details, there is unfortunately no easy getcap (how to retrieve rsa keysizes?), just a tpm_testparms where you can 'test' whether the specified template is supported.
This is a whole new chunk of functionality, which I then can plug in here.

For AES, yes correct, but if it's not supported at all, the mechanisms will not be in the mechanism_info_list. (All applications I'm aware of query the provider if a mechanissm is supported at all and then for the details. Or they don't query anything)

williamcroberts: I guess since it's only related to #198 and it leaves that open, it will serve as a reminder to ensure this doesn't last forever.

PR #252, opened 2019-06-04
Add DB schema for future upgrades

### LINKED ISSUE #195: db: add schema upgrade support
add a schema table with 1 row and column that has the schema version so we can upgrade the db and deal with different formats.

PeterHuewe:
We need a way to track the current version of our db format.
If the format (e.g. table structure or names) change, an update path/tool
can be created based on the schema version.

Fixes #195

### DIFF
--- a/tools/tpm2_pkcs11/db.py
+++ b/tools/tpm2_pkcs11/db.py
@@ -295,6 +295,7 @@ class Db(object):
     # secondary
     # tertiary
     #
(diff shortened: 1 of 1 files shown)


PeterHuewe:
Build fails due to some unknown reason in travis ci, regardless of the change itself (see master is failing).
tpm2-tools or tpm2-tss update probably broke something.

williamcroberts:
build restarted

Issue #232, opened 2019-05-22
Causes stack smashing crash on firefox and with nss modutil


bepvte:
I tried to make a recording of how I encounter the issue: https://asciinema.org/a/247656

I could upload the coredumps, but I suspect they would not provide much help without some kind of debugging symbols or something. Let me know if theres anything I can provide.

When adding in firefox, a crash happens every time you try to check your pkcs11 device list, even in safe mode. The solution is to delete the pkcs11.txt in your firefox profile, if anyone else has this problem.
Here is the firefox crash report: https://crash-stats.mozilla.org/report/index/31527d0f-0bfd-46c7-a4c3-d45240190522

PeterHuewe:
Hi,
unfortunately I cannot reproduce it on my raspberry pi system.
```
$ mkdir -p $HOME/.pki/nssdb
$ certutil -d $HOME/.pki/nssdb -N
Enter a password which will be used to encrypt your keys.
The password should be at least 8 characters long,
and should contain at least one non-alphabetic character.

Enter new password: 
Re-enter password: 

$ modutil -dbdir sql:/home/pi/.pki/nssdb/ -add "tpm2" -libfile /usr/lib/arm-linux-gnueabihf/pkcs11/libtpm2_pkcs11.so

WARNING: Performing this operation while the browser is running could cause
corruption of your security databases. If the browser is currently running,
you should exit browser before continuing this operation. Type 
'q <enter>' to abort, or <enter> to continue: 

Module "tpm2" added to database.
```

Any ideas how I can reproduce this?

Maybe you could run
`TPM2_PKCS11_LOG_LEVEL=2 modutil ....`


Can you specify what you mean by deleting the pkcs11.txt ? How does that solve the problem?

redg3ar:
It doesnt seem to occur when running a build with no or low optimization (-Og or -O0).
Output of the increased loglevel below:
```
INFO on line: "395" in file: "src/pkcs11.c": enter "C_GetFunctionList"
INFO on line: "395" in file: "src/pkcs11.c": return "C_GetFunctionList" value: 0
INFO on line: "383" in file: "src/pkcs11.c": enter "C_Initialize"
INFO on line: "1903" in file: "src/lib/db.c": Using sqlite3 DB: "/home/bep/tmp/tpm2_pkcs11.sqlite3"
INFO on line: "383" in file: "src/pkcs11.c": return "C_Initialize" value: 0
[...]

PR #255, opened 2019-06-07
utils: replace min(strlen(s),l) by strnlen(s,l) in str_padded_copy

diabonas:
`str_padded_copy` is used both for zero-terminated strings as well as fixed char arrays (blank padded). `strlen` doesn't work for arrays without zero termination and does an out-of-bounds read until it finds a null character. This triggers stack smashing protection resulting in a segfault on some systems. Using `strnlen(s,l)` is functionally equivalent and guarantees that at most l characters are consumed, avoiding a buffer over-read.

Fixes: #232

### DIFF
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -24,14 +24,6 @@
 
 #define UNUSED(x) (void)x
 
-static inline int max(size_t a, size_t b) {
-    return a > b ? a : b;
-}
-
-static inline int min(size_t a, size_t b) {
-    return a < b ? a : b;
-}
-
 /*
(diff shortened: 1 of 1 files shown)


diabonas:
Whops, I think I lied a bit in the commit message: it's not the reading itself that triggers the stack smashing protection, apparently the `min()` function doesn't actually return the minimum sometime, so this is really a buffer overflow. Very weird, the function itself looks ok, apart from the return type, but changing `int` to `size_t` doesn't help either. Maybe some kind of comparison problem related to type conversion or some `inline` weirdness?

Anyway, I updated the commit message and removed the apparently broken `min()` and `max()` functions, the solution of using `strnlen` stays the same.

Issue #199, opened 2019-04-11
C_GenerateKeyPair (ECC): Check if curve is supported by the TPM


PeterHuewe:
My TPM only supports 
```
tpm2_getcap -c ecc-curves
TPM2_ECC_NIST_P256: 0x3
TPM2_ECC_BN_P256: 0x10
```
Thus if I select a P224 curve, the current code fails quite violently.
```
13: C_GenerateKeyPair
2019-04-11 17:28:29.503
[in] hSession = 0x1000001
pMechanism->type=CKM_EC_KEY_PAIR_GEN          
[in] pPublicKeyTemplate[6]: 
    CKA_CLASS             CKO_PUBLIC_KEY       
    CKA_TOKEN             True
    CKA_VERIFY            True
    CKA_LABEL             00fef8f8 / 5
    4C5A454D 6B
     L Z E M  k
    CKA_KEY_TYPE          CKK_EC             
    CKA_ECDSA_PARAMS      00ff2f58 / 7
    00000000  06 05 2B 81 04 00 21                             ..+...!         
[in] pPrivateKeyTemplate[5]: 
    CKA_TOKEN             True
    CKA_SIGN              True
    CKA_LABEL             0101d520 / 5
    4C5A454D 6B
     L Z E M  k
    CKA_SENSITIVE         True
    CKA_EXTRACTABLE       False
INFO on line: "624" in file: "src/pkcs11.c": enter "C_GenerateKeyPair"
WARNING:esys:src/tss2-esys/api/Esys_CreateLoaded.c:360:Esys_CreateLoaded_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_CreateLoaded.c:127:Esys_CreateLoaded() Esys Finish ErrorCode (0x000002e6) 
ERROR on line: "2165" in file: "src/lib/tpm.c": Esys_CreateLoaded: 0x2e6
ERROR on line: "2548" in file: "src/lib/tpm.c": create_loaded 0x2e6
ERROR on line: "558" in file: "src/lib/key.c": Failed to generate key
INFO on line: "624" in file: "src/pkcs11.c": return "C_GenerateKeyPair" value: 5
Returned:  5 CKR_GENERAL_ERROR
```

--> we need to add a check for whether the curve is supported by the TPM.

williamcroberts:
Isn't this handled by #198 

[...]

PR #256, opened 2019-06-12
Add check for tpm ecc-capability to avoid tpm-error

StefanieLautenbacher:
Get supported ecc curves from tpm and check, if the chosen curve is supported,
before key generation in tpm.

Signed-off-by: Stefanie Lautenbacher <Stefanie.Lautenbacher@infineon.com>

### DIFF
--- a/src/lib/tpm.c
+++ b/src/lib/tpm.c
@@ -2478,6 +2478,51 @@ out:
     return rv;
 }
 
+static CK_RV check_tpm_ecc(ESYS_CONTEXT *context, tpm_key_data tpmdat) {
+
+    /* get ecc-curve capability from TPM*/
+    TPM2_CAP capability = TPM2_CAP_ECC_CURVES;
+    UINT32 property = TPM2_ECC_NIST_P192;
+    UINT32 counter = TPM2_MAX_ECC_CURVES;
+    TPMI_YES_NO more_data;
+    TPMS_CAPABILITY_DATA *capabilityData;
+
+    TSS2_RC rval = Esys_GetCapability (context,
+            ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE,
+            capability, property, counter,
+            &more_data, &capabilityData);
+
+    if (rval != TSS2_RC_SUCCESS) {
+        LOGE("Esys_GetCapability: 0x%x:", rval);
+        return CKR_GENERAL_ERROR;
+    }
+
+    if (!capabilityData || capabilityData->capability != capability) {
+        LOGE("TPM returned different capability than requested: 0x%x != 0x%x",
+            capabilityData->capability, capability);
+        rval = CKR_GENERAL_ERROR;
+        goto out;
+    }
+
+    TPMU_CAPABILITIES *capabilities = &capabilityData->data;
+
+    /* Compare the requested ecc curve to the supported curves*/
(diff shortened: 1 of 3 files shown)


PeterHuewe:
There's a bug in tpm.c line
```
 ../src/lib/tpm.c:2502:13: warning: Access to field 'capability' results in a dereference of a null pointer (loaded from variable 'capabilityData')
            capabilityData->capability, capability);
```
If !capabilityData is true, you mustn't print it

src/lib/tpm.c
```diff
@@ -2538,6 +2583,16 @@ CK_RV tpm2_generate_key(
     auth->size = len;
     memcpy(auth->buffer, newauthbin, auth->size);
 
+    /*
+     * Check the ecc support of the TPM
+     */
+    if (mechanism->mechanism == CKM_EC_KEY_PAIR_GEN) {
+        rv = check_tpm_ecc(tpm->esys_ctx, tpmdat);
```
williamcroberts: Doesn't the RC from create let you know if the curve isn't supported? Wouldn't this be some type of specific PKCS11 error code that should be set here?

src/lib/tpm.c
```diff
@@ -2478,6 +2478,51 @@ static CK_RV tpm_object_data_populate_ecc(TPM2B_PUBLIC *out_pub, tpm_object_data
...
+    UINT32 property = TPM2_ECC_NIST_P192;
+    UINT32 counter = TPM2_MAX_ECC_CURVES;
+    TPMI_YES_NO more_data;
+    TPMS_CAPABILITY_DATA *capabilityData;
+
+    TSS2_RC rval = Esys_GetCapability (context,
+            ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE,
+            capability, property, counter,
+            &more_data, &capabilityData);
+
```
williamcroberts: no space between routine and checking conditional.

src/lib/tpm.c
```diff
@@ -2478,6 +2478,51 @@ static CK_RV tpm_object_data_populate_ecc(TPM2B_PUBLIC *out_pub, tpm_object_data
...
+    if (rval != TSS2_RC_SUCCESS) {
+        LOGE("Esys_GetCapability: 0x%x:", rval);
+        return CKR_GENERAL_ERROR;
+    }
+
+    if (!capabilityData || capabilityData->capability != capability) {
```
williamcroberts: FYI: if rval == TSS2_SUCCESS then capabilityData is guaranteed NOT to be NULL. This is the static analysis tool getting it wrong. I would say just drop an assert(capabilityData) after the check of rval..

test/integration/pkcs-tpm-ecc-support-check.int.c
```diff
@@ -0,0 +1,132 @@
...
+            /*P521 is not supported*/
+            assert_int_equal(rv, CKR_ARGUMENTS_BAD);;
+        }
+    }
+}
+
+int main() {
+
+    const struct CMUnitTest tests[] = {
+        cmocka_unit_test_setup_teardown(test_tpm_ecc_support_check,
```
williamcroberts: Wouldn't this just be in the keygen test file that already exists?

src/lib/tpm.c
PeterHuewe: Basically what happens is #199
You get some not nice error messages in the log, instead of gracefully telling the user his params are wrong.

Issue #238, opened 2019-05-28
Calling C_Sign for size prevents C_DestroyObject


PeterHuewe:
The tobj->active counter is increased by a call to C_Sign, but if C_Sign is just called for size (pSignature=NULL), it never decreases.

For easier understanding I annotated the functions in object.c:
```
--- a/src/lib/object.c
+++ b/src/lib/object.c
@@ -568,6 +568,7 @@ CK_RV tobject_user_increment(tobject *tobj) {
     }
 
     tobj->active++;
+    LOGE("Active Value INC: %d", tobj->active);
 
     return CKR_OK;
 }
@@ -580,12 +581,14 @@ CK_RV tobject_user_decrement(tobject *tobj) {
     }
 
     tobj->active--;
+    LOGE("Active Value DEC: %d", tobj->active);
 
     return CKR_OK;
 }
 
 static bool tobject_is_busy(tobject *tobj) {
     assert(tobj);
+    LOGE("Active Value: %d", tobj->active);
 
     return tobj->active > 0;
 }
```

And get the following result:
```
24: C_SignInit
2019-05-28 17:05:11.270
[in] hSession = 0x1000000
pMechanism->type=CKM_RSA_PKCS                 
[in] hKey = 0x9
ERROR: Active Value INC: 1
Returned:  0 CKR_OK

25: C_Sign
2019-05-28 17:05:11.270
[in] hSession = 0x1000000
[in] pData[ulDataLen] 1231c600 / 51
    00000000  30 31 30 0D 06 09 60 86 48 01 65 03 04 02 01 05  010...`.H.e.....
    00000010  00 04 20 29 FF 07 F0 8F 0A 74 38 2E 62 56 C4 6A  .. ).....t8.bV.j
    00000020  54 3E C0 38 D0 C3 8B 15 13 60 5B 33 A2 85 82 A4  T>.8.....`[3....
    00000030  89 FA 1E                                         ...             
ERROR: Active Value INC: 2
[out] pSignature[*pulSignatureLen] NULL [size : 0x100 (256)]
Returned:  0 CKR_OK

26: C_Sign
2019-05-28 17:05:11.575
[in] hSession = 0x1000000
[in] pData[ulDataLen] 1231c600 / 51
    00000000  30 31 30 0D 06 09 60 86 48 01 65 03 04 02 01 05  010...`.H.e.....
[...]

PR #257, opened 2019-06-13
Fix reference counting for CKM_RSA_PKCS

PeterHuewe:
In the special case of CKM_RSA_PKCS and other 'raw' hashes the sign operation
has to call the tpm for a decrypt which increases the internal reference
counters. Unfortunately these counters were not reset correctly.
Without breaking the existing logic and reworking everything,
the sign operation has to take care that the counters are decremented.

Fixes: #238


PeterHuewe:
Build is broken (again) due to update in tpm2_tools (again)

PeterHuewe:
Build fixed in #259

PR #262, opened 2019-07-05
tpm2_ptool: update tpm2-tools option from -b to -a

diabonas:
This has been changed upstream in https://github.com/tpm2-software/tpm2-tools/pull/1582, cf. the currently [failing build](https://travis-ci.org/tpm2-software/tpm2-pkcs11/jobs/554728372#L2746).


williamcroberts:
Thank you!

Issue #260, opened 2019-06-20
drop secondary object


williamcroberts:
Just put all objects under the primary object... this *i think* will make our data model match the OSSL engine. Thus, we can use our keys with the OSSL engine

PR #265, opened 2019-07-10
doc: Add architecture description

AndreasFuchsSIT:
I tried to put my grasp of the inner workings into words.
Please tell me, if I made any mistakes.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1,0 +1,62 @@
+The purpose of this file is to describe the functional and key architecture of
+this project. It is spiced with some rational throughout.
+
+# Slots & Tokens
+
+This library shall provide multiple tokens for the user(s). Any given slot can
+only contain one token at any given time. In case of smartcards the slot is a
+reader whilst the token is the actual smartcard. Since this project won't
+implement a separate tool for "virtual swapping" of tokens, each token is
+associated with its own slot. Thus, the library offers a set of slots with a
+token inside each of these.
+
+In order to allow the creation of new tokens using the pkcs11 means, this
+library presents one more slot/token pair than tokens currently exist. The
+additional token is a blank token that can be initialized using the C_InitToken
+function.
+(Note: This is not implemented yet)
+
+# Objects
+
+Each token can contain an arbitrary amount of keys and other object. (Note:
+other objects are not implemented yet). The keys are TPM-based keys, i.e. they
+are only known to the TPM and also the key operations are executed inside the
+TPM. The other objects are encrypted using hybrid encryption with the outer
+key being sealed to the TPM.
+
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to delete them, if your attacker model allows access to
+the HDD.
+
+# Key Authorizations
+
+For each token, two TPM seal blobs exist one for the SO and one for the User.
+The authValue for these seal blobs are SO's PIN and the User's PIN. The data
+sealed inside is an intermediate key that is used to decrypt a second authValue
+stored on disk.
+
+This second authValue is then used to authenticate the use of
+the key objects. In case of a PIN-change only the seal blobs need to be changed,
+which is an atomic operation.
+
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to invalidate the copy of the blob with the old PIN, if
+your attacker model allows access to the HDD.
+
+# Key Hierarchies
+
+All keys reside under the so-called SRK. This is an (optionally persistent)
+primary key inside the TPM as specified by the TCG Provisioning Guidance.
+
+Note: Currently there is an additional intermediate storage per token that will
+be removed.
+
+# Storage
+
+All TPM-encrypted key blobs, the meta-data and the encrypted second authValue
+are stored in the OS-user's personal storage (i.e. HOME directory).
+
+
+
+
+


docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+reader whilst the token is the actual smartcard. Since this project won't
+implement a separate tool for "virtual swapping" of tokens, each token is
+associated with its own slot. Thus, the library offers a set of slots with a
+token inside each of these.
+
+In order to allow the creation of new tokens using the pkcs11 means, this
+library presents one more slot/token pair than tokens currently exist. The
+additional token is a blank token that can be initialized using the C_InitToken
+function.
+(Note: This is not implemented yet)
```
williamcroberts: No that's how SoftHSM works. You can use the ptool command "addtoken" to add an uninitialized token if you wanted too.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Objects
+
+Each token can contain an arbitrary amount of keys and other object. (Note:
+other objects are not implemented yet). The keys are TPM-based keys, i.e. they
+are only known to the TPM and also the key operations are executed inside the
+TPM. The other objects are encrypted using hybrid encryption with the outer
+key being sealed to the TPM.
```
williamcroberts: depends, small ones would be sealed to the TPM big ones use wrapping with a tpm key. I guess that probably doesn't matter and we could always just create key object + wrap.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to delete them, if your attacker model allows access to
+the HDD.
+
+# Key Authorizations
+
+For each token, two TPM seal blobs exist one for the SO and one for the User.
+The authValue for these seal blobs are SO's PIN and the User's PIN. The data
```
williamcroberts: yes, but:
  - respective to the so and user seal objects.
  - the pins are fed through a pbkdf and the output of that is the seal object authvalue.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+sealed inside is an intermediate key that is used to decrypt a second authValue
+stored on disk.
```
williamcroberts: sealed inside is a wrapping key used to encrypt other auth values OR an auth value to a tpm key. The goal of having the tpm approach is that if someone accesses the disk meta data, brute forcing the TPM key auth value would trigger DA protections. Although If its wrapped with AES 256 or 128, you would be bruteforcing 128 or 256 bits of entropy, which is less then the TPM key authvalue itself. So you wouldn't try that, you would just keep trying the data itself. I think I am open to just doing this in software.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+
+This second authValue is then used to authenticate the use of
```
williamcroberts: This intermediate key is used to unwrap the other objects authvalues.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to invalidate the copy of the blob with the old PIN, if
+your attacker model allows access to the HDD.
+
+# Key Hierarchies
+
+All keys reside under the so-called SRK. This is an (optionally persistent)
+primary key inside the TPM as specified by the TCG Provisioning Guidance.
+
+Note: Currently there is an additional intermediate storage per token that will
```
williamcroberts: none as the sobject, I would plop that in there so we don't get confused what to remove later. It's also on this bug marked for alpha: https://github.com/tpm2-software/tpm2-pkcs11/issues/260

williamcroberts:
This whole things is part of #8

docs/ARCHITECTURE.md
AndreasFuchsSIT: That was my thinking on how to do things after "rearranging" some of the code.
Nice this would be that we then don't need any external tools anymore.
Or would you say that such an empty token would be too confusing and we should only allow adding of tokens via an extra command ?

docs/ARCHITECTURE.md
AndreasFuchsSIT: I guess that's true. Though it makes me think if a single code-path would be more beneficial than saving the intermediate key for small payloads...
What do you think ?

docs/ARCHITECTURE.md
AndreasFuchsSIT: Yes, I actually did not include the TPM2_EncrytDescrypt2() usage in this description. I was assuming software here... ;-)

docs/ARCHITECTURE.md
AndreasFuchsSIT: I don't think I get what you're saying ?

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+which is an atomic operation.
+
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to invalidate the copy of the blob with the old PIN, if
+your attacker model allows access to the HDD.
+
+# Key Hierarchies
+
+All keys reside under the so-called SRK. This is an (optionally persistent)
+primary key inside the TPM as specified by the TCG Provisioning Guidance.
```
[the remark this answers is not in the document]
AndreasFuchsSIT: Ok... Do we want to keep this primary key configurable going forward or do we want to restrict it to the SRK ? (I'd be in favor of the latter)

AndreasFuchsSIT:
Thanks for the review...
I'll update the doc... Some uestions posted above...
(why is my lowercase Q key not working in firefox ???)

docs/ARCHITECTURE.md
williamcroberts: I don't think it's super confusing, as SoftHSM does it. But I think it's odd. As you create tokens a new one keeps existing.

I don't think you ever get away from external commands. The PKCS#11 spec isn't 100% functional for card initialization. Offhand I forget where it gets weird, but everything always ships a provisioning tool, even SoftHSM

docs/ARCHITECTURE.md
williamcroberts: I don't know how you would get to a single code path unless that code path was just "create key and encrypt blob".

docs/ARCHITECTURE.md
williamcroberts: I think we want to keep TPM based wrapping if possible, if only for a small subset and for DA protections and rate limiting protections on that key.

docs/ARCHITECTURE.md
AndreasFuchsSIT: I guess creating a token with an SO-PIN only can be done with a single future-api call. Adding the user-PIN via the SO-PIN and then creating the key object would then be, where PKCS11 functions take over ?

docs/ARCHITECTURE.md
williamcroberts: The intermediate storage token is internally referenced and known as the "secondary object" or sobject. So place something like:

```md
Note: Currently theere is an additional intermediate storage object (sobject) per token that will...
```

docs/ARCHITECTURE.md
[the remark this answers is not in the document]
williamcroberts: Configurable, a lot of IoT devices in the wild are not set up properly and we need to support them as much as possible.

docs/ARCHITECTURE.md
AndreasFuchsSIT: Well, instead of going
a) Create intermediate key and seal that and encrypt with that AND
b) Don't create intermediate keys but seal data directly
We could'd just get rid of b) and always create an intermediate key. That's the reduced code-path I'm referring to...
What do you think ?

docs/ARCHITECTURE.md
williamcroberts: @AndreasFuchsSIT I think they may actually just be limitations on the actual card side

docs/ARCHITECTURE.md
williamcroberts: Yeah, we're thinking the same here.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+
+Each token can contain an arbitrary amount of keys and other object. (Note:
+other objects are not implemented yet). The keys are TPM-based keys, i.e. they
+are only known to the TPM and also the key operations are executed inside the
+TPM. The other objects are encrypted using hybrid encryption with the outer
+key being sealed to the TPM.
+
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to delete them, if your attacker model allows access to
+the HDD.
```
mdempsky: I'm not sure what this paragraph is trying to say. I think it's warning that if you delete the HDD-stored objects, they'll be lost? And that if you allow attackers to access your file system, they might be able to delete them?

If so, I think this could be more clearly worded as something like:

> Note: TPMs have very limited memory. To support an arbitrary number of objects, objects are stored (encrypted) in the file system and loaded/unloaded into the TPM as needed. If objects are deleted from the file system, they may be lost.

williamcroberts: It also means that if an attacker ever had read access to this location, when you delete it, it doesn't matter, because they have a copy of the object.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+
+In order to allow the creation of new tokens using the pkcs11 means, this
+library presents one more slot/token pair than tokens currently exist. The
+additional token is a blank token that can be initialized using the C_InitToken
+function.
+(Note: This is not implemented yet)
+
+# Objects
+
+Each token can contain an arbitrary amount of keys and other object. (Note:
```
mdempsky: Perhaps worth noting that in PKCS#11, all objects within a token share a PIN / login. So there's a difference between 1 token with N objects, and N tokens with 1 object each.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,62 @@
...
+TPM. The other objects are encrypted using hybrid encryption with the outer
+key being sealed to the TPM.
+
+Note: Since the TPM's object are stored (encrypted) external to the TPM on the
+HDD, it is not possible to delete them, if your attacker model allows access to
+the HDD.
+
+# Key Authorizations
+
+For each token, two TPM seal blobs exist one for the SO and one for the User.
```
mdempsky: Why not use PIN Fail NV indices, and key policies that reference them?

Also, what's the point of the SO PIN here? Seems redundant when there's already owner authorization over the TPM.

docs/ARCHITECTURE.md
mdempsky: Does using PBKDF add anything here? I'm not strictly opposed to using it, but as far as I can tell, it only offers two features:

1. It's slow, which helps to slow down brute force; but that should be handled anyway by the TPM's DA protection (or using PIN Fail NV Indices).

2. It compresses long PINs so that they're smaller than the key's NameAlg hash output size (as auth values need to be). But this could be handled by a single hash iteration, and could be omitted for inputs that fit.

docs/ARCHITECTURE.md
williamcroberts: I wanted to avoid using any additional NV and use transient objects as much as possible. There also exists a goal to support password based objects, so existing objects can be migrated if the attributes allow for it and objects can be easily migrated out if needed. Most objects currently in use just have a password. With that said, the SOPIN allows the SO to reset the user password.

You're not guaranteed that you have owner authorization over the TPM. You have access to a persistent primary object, which could be under any hierarchy. Currently the primary object is rooted in the owner hierarchy, but you never grant owner hierarchy permissions to tokens/so. However, i'd like the hierarchy to be configurable at setup time.

docs/ARCHITECTURE.md
williamcroberts: Really just 1 because as you point out 2. It made a little more sense when the db and files were separate because it made the attacker need both the file and db, but now it just requires the db. So in the off chance someone did something sloppy it threw a chair in the hallway.

docs/ARCHITECTURE.md
mdempsky: I see. I think maybe it would be helpful if the document somewhere specified the use cases and constraints it's targeting.

E.g., I'm interested in using tpm2-pkcs11 on my own personal devices, where I'm the TPM2 owner, and where the entire reason I'm using it is to be able to prevent PIN brute forcing and key ex-filtration. I understand TPM2 supports more complicated ownership models for enterprises, but I'm not personally familiar with them, and it's not been a priority for me to learn them yet.

I also expect that if I'm concerned that my PIN may have been compromised, that I can just change it and feel confident the old PINs will no longer work. My understanding of TPM2 is this sort of revocation can't happen without involving NV. (Maybe clock-based policies could provide some safety guarantee that old PINs eventually expire? But you'd have to re-seal the auth value periodically or risk getting locked out.)

I can appreciate that there's limited NV space on a TPM2, and so it's probably best to minimize how many indices are used anyway (e.g., recommending to group multiple keys onto a single "token" so they can share a single PIN / NV Index). But I think the default should be to use NV space; and for users who can't or don't want to use NV, to support a separate less secure mode that uses only transient objects (and has to be explicitly enabled, so they're aware of the reduced security).

williamcroberts: > I see. I think maybe it would be helpful if the document somewhere specified the use cases and constraints it's targeting.

My goal with the "document architecture doc" was to do that. Andreas was nice enough to take a stab at it so he understood the constraints as hes working on an internals change as we speak.

> 
> E.g., I'm interested in using tpm2-pkcs11 on my own personal devices, where I'm the TPM2 owner, and where the entire reason I'm using it is to be able to prevent PIN brute forcing and key ex-filtration. I understand TPM2 supports more complicated ownership models for enterprises, but I'm not personally familiar with them, and it's not been a priority for me to learn them yet.

This is unfortunately where technology meets people and weirdness ensues.

> 
> I also expect that if I'm concerned that my PIN may have been compromised, that I can just change it and feel confident the old PINs will no longer work. My understanding of TPM2 is this sort of revocation can't happen without involving NV. (Maybe clock-based policies could provide some safety guarantee that old PINs eventually expire? But you'd have to re-seal the auth value periodically or risk getting locked out.)
> 
> I can appreciate that there's limited NV space on a TPM2, and so it's probably best to minimize how many indices are used anyway (e.g., recommending to group multiple keys onto a single "token" so they can share a single PIN / NV Index). But I think the default should be to use NV space; and for users who can't or don't want to use NV, to support a separate less secure mode that uses only transient objects (and has to be explicitly enabled, so they're aware of the reduced security).

We're actually working up a new authmodel design that policy based and supports wrapping auth values too. We plan on writing that up and sharing it next week for you and others to comment on so we can incorporate it with the internals change. @idesai in the new model, on the policy side, if the password changes does it invalidate old pins?

idesai:
PR #294 replaces this.

PR #266, opened 2019-07-10
Test: Run tests without setup script

AndreasFuchsSIT:
Fix a bug where tests would not run unless a setup script was provided.

### DIFF
--- a/test/integration/scripts/int-test-setup.sh
+++ b/test/integration/scripts/int-test-setup.sh
@@ -204,6 +204,10 @@ echo ${TPM2_PKCS11_TCTI}
 # if provided, run the test script
 if [ -z "${TSETUP_SCRIPT}" ]; then
     echo "No setup script provided"
+    # execute the test script and capture exit code
+    echo $@
+    env $@
+    ret_test=$?
 else
 	echo "Run setup script ${TSETUP_SCRIPT}"
     sh ${TSETUP_SCRIPT} --tmpdir=${SIM_TMP_DIR}


williamcroberts:
I don't understand the problem or how that change fixes it, can you elaborate please?

test/integration/scripts/int-test-setup.sh
```diff
@@ -204,6 +204,10 @@ echo ${TPM2_PKCS11_TCTI}
 # if provided, run the test script
 if [ -z "${TSETUP_SCRIPT}" ]; then
```
williamcroberts: We could simplify and reduce the duplication o
```
if -n TEST_SETUP_SCRIPT:
    then run setup script
fi

run test

AndreasFuchsSIT:
> I don't understand the problem or how that change fixes it, can you elaborate please?

The ``--tsetup-script`` parameter is options, but if left out, the test is not run.

So either ``--tsetup-script`` should be mandatory and checked or (as I assumed) it's actually optional and tests should then be run.

I can deduplicate the test invocation if you like, shall I ?

williamcroberts:
> > I don't understand the problem or how that change fixes it, can you elaborate please?
> 
> The `--tsetup-script` parameter is options, but if left out, the test is not run.
> 
> So either `--tsetup-script` should be mandatory and checked or (as I assumed) it's actually optional and tests should then be run.
> 
> I can deduplicate the test invocation if you like, shall I ?

yeah, I get the problem now. Please de-dup and simplify the code paths.

PR #269, opened 2019-07-16
tests: Use tpm2-pyesys for simulator testing

pdxjohnny:
This is my WIP branch for python binding integration

### DIFF
--- a/tools/tests/test_ptool.py
+++ b/tools/tests/test_ptool.py
@@ -1,0 +1,45 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2019 Intel Corporation
+# All rights reserved.
+import sys
+import unittest
+import subprocess
+from unittest.mock import patch
+
+from tpm2_pyesys.util.simulator import SimulatorTest
+
+from tpm2_pkcs11 import tpm2_ptool as tool
+from tpm2_pkcs11.utils import TemporaryDirectory
+
+
+class TestPTool(SimulatorTest):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.abrmd = subprocess.Popen(
+            ["sudo", "-u", "tss", "tpm2-abrmd", "--tcti=mssim"]
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        for pid in [
+            line.split()[1]
+            for line in subprocess.check_output(["ps", "aux"]).decode().split("\n")
+            if "tpm2-abrmd" in line
+        ]:
+            subprocess.call(["sudo", "kill", pid])
+        super().tearDownClass()
+        cls.abrmd.wait()
+
+    def test_init(self):
+        with TemporaryDirectory() as tempdir:
+            with patch("sys.exit"), patch(
+                "sys.argv",
+                new=[
+                    "tpm2_ptool.py",
+                    "init",
+                    "--pobj-pin=mypobjpin",
+                    "--path=%s" % (tempdir,),
+                ],
+            ):
+                tool.main()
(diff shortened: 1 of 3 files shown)


tools/tests/test_ptool.py
```diff
@@ -0,0 +1,45 @@
...
+from tpm2_pkcs11 import tpm2_ptool as tool
+from tpm2_pkcs11.utils import TemporaryDirectory
+
+
+class TestPTool(SimulatorTest):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.abrmd = subprocess.Popen(
+            ["sudo", "-u", "tss", "tpm2-abrmd", "--tcti=mssim"]
```
pdxjohnny: This is super not ideal, I looked at the abrmd repo and am very unclear on what's going on there. Would fixing https://github.com/tpm2-software/tpm2-abrmd/issues/139 make `abrmd` easier to run in a test environment like this?

My code to start the simulator is essentially `tpm_server -rm` followed by connect calls until one of them succeeds then we know it has bound and `SimulatorTest.setUpClass` returns.

tools/tests/test_ptool.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        for pid in [
+            line.split()[1]
+            for line in subprocess.check_output(["ps", "aux"]).decode().split("\n")
+            if "tpm2-abrmd" in line
+        ]:
+            subprocess.call(["sudo", "kill", pid])
```
pdxjohnny: I hate myself for this but its the only way obi-wan

pdxjohnny: I'm assuming that having to run as root or tss user has to do with dbus, or is that incorrect.

Ping: @flihp @AndreasFuchsSIT

tools/tests/test_ptool.py
```diff
@@ -0,0 +1,45 @@
...
+        super().tearDownClass()
+        cls.abrmd.wait()
+
+    def test_init(self):
+        with TemporaryDirectory() as tempdir:
+            with patch("sys.exit"), patch(
```
pdxjohnny: Oops I shouldn't be patching `sys.exit`

tools/tests/test_ptool.py
AndreasFuchsSIT: tpm2-abrmd needs to run as user tss/root, because /dev/tpm0 is tss:root and /dev/tpmrm0 is tss:tss.
If you're using mssim, you can however run tpm2-abrmd as $USER without sudo...

P.S. I'm reworking tpm2-pkcs11 atm which may void python-code in this project. Just as a heads-up...
tpm2-pyesys remains valid though...

pdxjohnny: Ah okay great to know. Thank you!

tools/tests/test_ptool.py
williamcroberts: You can run it in a way where your no on the session bus and then you don't need higher permissions. Look at how we set up the make check tests. We get a simulator and abrmd fired up in this project.

tools/tests/test_ptool.py
williamcroberts: Step 1 is to not care about using an actual TPM device and just build all your testing for the simulator.

williamcroberts:
our makefile check framework fires all this stuff up, can you just add it as an integration test? Do we need python binding tests, the current command line calls just get tested in the integration tests... isn't that enough?

pdxjohnny:
When I put this up it was based on an early version of the bindings. Many
things changed since then so I think yes we can use the existing harness
stuff

pdxjohnny:
Will re-open tpm2-pytss support via new branch soon

PR #274, opened 2019-07-24
Fix change of tool options

joholl:
Some tool options changed and thus have to be changed here, too.

- tpm2_createprimary: `-o` became `-c` (tpm2-software/tpm2-tools@3939d715)
- tpm2_load: `-o` became `-c` (tpm2-software/tpm2-tools@888f63c4)

Additionally, a typo was fixed.


williamcroberts:
I think well see this die on a few others too.

PR #293, opened 2019-08-28
tpm2_ptool: move 'noqa' annotation to end of line

diabonas:
lgtm.com does not recognise the suppression comment [if there is trailing text after it](https://help.semmle.com/lgtm-enterprise/user/help/alert-suppression.html#format), while Pylint does not mind a trailing comment, so move `noqa` to the end to appease both Pylint and LGTM.


diabonas:
I am excited to see the new version :tada: No worries, these were the last alerts to be fixed/suppressed, I just like to keep the number of warnings as close to zero as possible to make it easier to spot new problems :)

PR #296, opened 2019-10-10
WIP: FAPI

AndreasFuchsSIT:
This is a wip status of the port from ESYS to FAPI.
The primary goal is to get rid of the python tool and sqlite database.
Instead we're now relying on the FAPI built-in keystore.
Also, we can utilize FAPI policies in the future to come up with more interesting
schemes.

The architecture has two seal tokens (one for SO and one for USER) that contain the
same intermediate secret. That secret is used to authenticate against all of the other
key objects stored in the TSS keystore.

The integration tests have been shifted to a bash-script running pkcs11-tool. I also want
to add p11tool and openssl-pcks11-engine based test scripts.
The old integration tests are disabled, because the assume 3 active tokens at startup.
So they need to be adapted.

This is a very early stage; missing ECDSA, PSS and OAEP. PSS and OAEP should work and I have
not idea on why it does not work. Also we're missing custom key sizes, which is something
we have to discuss in FAPI; do we allow this or do we consider this a security policy breach ?
Also missing a lot of mechanism and attribute handling all over the place and still carrying
a lot of dead code from before.

I'll be force-pushing to this PR and trying to update the PR message as good as possible
until we can merge this into master and work in there. I'd give it 1-2 more weeks.

CI with the reduced integration tests may even work; make check does...


src/lib/tpm.c
```diff
@@ -29,339 +30,208 @@
...
-}
+        if (!slot_list)
+            continue;
 
-static void flags_restore(tpm_ctx *ctx) {
+        if (numPaths > *count) {
+            LOGE("buffer too small, count = %u, required=%u", *count, numPaths);
+            Fapi_Free(pathlist);
+            return CKR_BUFFER_TOO_SMALL;
+        }
```
arsing: This has the same bug as https://github.com/tpm2-software/tpm2-pkcs11/issues/299

AndreasFuchsSIT: fixed

williamcroberts:
Unfortunately, the FAPI doesn't expose all the parts of the API we need to implement a fully functional PKCS11 implementation. For instance, specifying key size or getting the key component material.

PR #305, opened 2019-11-07
DO NOT MERGE TESTING CI

williamcroberts:
We should see CI fail with:

src/lib/general.c: In function ‘unused_function’:
src/lib/general.c:54:26: error: unused parameter ‘x’ [-Werror=unused-parameter]
 void unused_function(int x) {
                          ^


williamcroberts:
Fixed: 5bae19635639f8

Issue #301, opened 2019-11-02
implement new authmodel using policies


williamcroberts:
Implement your auth model

idesai:
- [ ] Change authentications to policy type.
- [ ] Add policy hash data to the schema.
- [ ] Change object types created explicitly under tpm2-pkcs11 to be policy authorized.
- [ ] Restrict primary object to either Load/ Import for external objects and create for objects explicitly created under tpm2-pkcs11.
- [ ] Assign NV Indexes for SO-PIN and USER-PIN authentication.
- [ ] Change the policy for SO-PIN/ USER-PIN/ USER objects.
- [ ] Drop salt/ hashes for SO-PIN/ USER-PIN/ USER objects from database (schema).

arturkow2000:
Hi, we want to use TPM for client authentication to remote server over TLS (mutual TLS), we want to bind the private key to PCR values. Is support for this feature planned in a near future? We are willing to help in implementation.
We though about implementing this implementing at least the part required to import an externally generated object and satisfy its policy at runtime. Would such a solution be accepted upstream?

williamcroberts:
We don't have any plans to implement it in the near future, but I would support taking patches. We could add a field to the object attributes, like we do for the TPM blobs that contains a JSON policy as understood by tpm2-tss libpolicy library. Then all you would need to do is call the policy library with the callbacks set to satisfy that policy. We would need options in tpm2_ptool to create such an object as well. Should be "fairly" straightforward and I can provide guidance along the way.

arturkow2000:
Thanks for a quick response. I took a brief look at libpolicy and I saw that it uses FAPI internally. We were thinking about using esysdb backend, because format database is more space-efficient (binary data instead JSON) and all data is kept in a single file which we can just grab and save to a separate partition, FAPI storage make this a little bit more complicated. Is it possible to add policy support to esysdb backend? If so, how hard would it be to implement policy support on Esys compared to FAPI+libpolicy?
If that turned to be to hard we would rethink our approach and switch to FAPI.

williamcroberts:
[...]

PR #308, opened 2019-11-07
Adding policy to primary object and adding policy information and field to the database

idesai:
Furthers #301

### DIFF
--- a/tools/tpm2_pkcs11/db.py
+++ b/tools/tpm2_pkcs11/db.py
@@ -243,6 +243,7 @@ class Db(object):
                 sopub BLOB NOT NULL,
                 sopriv BLOB NOT NULL,
                 soauthsalt TEXT NOT NULL,
+                sealobjauthpolicy BLOB,
                 FOREIGN KEY (tokid) REFERENCES tokens(id) ON DELETE CASCADE
             );
             '''),
@@ -251,7 +252,8 @@ class Db(object):
                 id INTEGER PRIMARY KEY,
                 hierarchy TEXT NOT NULL,
                 handle BLOB NOT NULL,
-                objauth TEXT NOT NULL
+                objauth TEXT NOT NULL,
+                pobjauthpolicy BLOB
             );
             '''),
             textwrap.dedent('''
@@ -263,6 +265,7 @@ class Db(object):
                 objauth TEXT NOT NULL,
                 mech TEXT NOT NULL,
                 attrs TEXT NOT NULL,
+                tobjauthpolicy BLOB,
                 FOREIGN KEY (tokid) REFERENCES tokens(id) ON DELETE CASCADE
             );
             '''),
--- a/tools/tpm2_pkcs11/db.py
+++ b/tools/tpm2_pkcs11/db.py
@@ -126,7 +126,7 @@ class Db(object):
 
         return c.lastrowid
 
-    def addprimary(self, tr_handle, objauth, hierarchy='o'):
+    def addprimary(self, tr_handle, objauth, policy, hierarchy='o'):
 
         # Subordiante commands will need some of this data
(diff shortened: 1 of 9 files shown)


tools/tpm2_pkcs11/db.py
```diff
@@ -240,7 +242,8 @@ def create(self):
                 id INTEGER PRIMARY KEY,
                 hierarchy TEXT NOT NULL,
                 handle BLOB NOT NULL,
-                objauth TEXT NOT NULL
+                objauth TEXT NOT NULL,
+                pobjauthpolicy BLOB
```
williamcroberts: primary objects have a policy hash? Why?

tools/tpm2_pkcs11/db.py
```diff
@@ -252,6 +255,7 @@ def create(self):
                 objauth TEXT NOT NULL,
                 mech TEXT NOT NULL,
                 attrs TEXT NOT NULL,
+                tobjauthpolicy BLOB,
```
williamcroberts: how can objauth remain non-null in the case of policy hash? We're storing the policy hash here right? If it's the same hash, why duplicate it throughout the whole db?

tools/tpm2_pkcs11/db.py
idesai: So that the only operations allowed are:
1. Import 
2. Load externally created objects
3. Create objects with the intended policy hash. This means that objects explicitly created with tpm2-pkcs11 are guaranteed to have a policy that accommodates the new auth model.

tools/tpm2_pkcs11/db.py
idesai: You are right. It cannot be a Null. I was phasing the changes in commits so I don't break the existing setup. So in the new auth model the concept of tobjects will vanish. In general though, that policypassword is a placeholder to bring other changes like policy based authorizations into the pkcs11 library. Once there, in the next phases/ commits the policy for the seal objects would be different from the primary objects.

williamcroberts: > You are right. It cannot be a Null. I was phasing the changes in commits so I don't break the existing setup. So in the new auth model the concept of tobjects will vanish. 

Why would the concept of tobjects vanish, they are the actual backing keys and objects the user interacts with via the pkcs11 API. Why would they go away?

> In general though, that policypassword is a placeholder to bring other changes like policy based authorizations into the pkcs11 library. Once there, in the next phases/ commits the policy for the seal objects would be different from the primary objects.

Their's two policies IIUC.  A policy for interacting with the "seal objects" and a policy for accessing the "tobjects". The soseal object would have 1 policy hash, the userseal object would have 1 policy hash and all the tobjects that use policy would share common hash.

williamcroberts: Anything that is shared in common in a db should be placed in a separate table with the value looked up by primary id. It helps prevent the db from bloating and normalizing databases.

idesai: > > You are right. It cannot be a Null. I was phasing the changes in commits so I don't break the existing setup. So in the new auth model the concept of tobjects will vanish.
> 
> Why would the concept of tobjects vanish, they are the actual backing keys and objects the user interacts with via the pkcs11 API. Why would they go away?
> 
> > In general though, that policypassword is a placeholder to bring other changes like policy based authorizations into the pkcs11 library. Once there, in the next phases/ commits the policy for the seal objects would be different from the primary objects.
> 
> Their's two policies IIUC. A policy for interacting with the "seal objects" and a policy for accessing the "tobjects". The soseal object would have 1 policy hash, the userseal object would have 1 policy hash and all the tobjects that use policy would share common hash.

That was my confusion. By tobjects I thought the parent objects were not primary objects which is why I thought that should vanish.

PR #324, opened 2019-11-21
Framework for policy based authorizations for SO/ USER role separation

idesai:
This commit:
1. Adds a policy store to the database that is initialized with preset
   policies.
2. The policy types is one of - NO_POLICY, PRIMARY, SOPIN, USERPIN,
   SEALOBJ, USEROBJ. Additionally it has listed standard policies like
   PolicyPassword. A combination of the standard policy types make
   the designated policies like PRIMARY, SOPIN, etc.
3. There is an option included to create objects without policies.
4. For objects without policies the object policy type field points to
   NO_POLICY.
5. It should be noted that the policy store remains for the life of the
   database file. The reason being, if no policies are needed, the object
   policytype field will need to point to NO_POLICY. If all the objects
   are deleted, it does not mean new objects cannot be created by issuing
   addkey commandlet. When this happens policies can be referenced from
   the database.
6. This commit adds policies for PRIMARY and USEROBJ only in spirit of
   putting a framework without breaking the test infrastructure. This
   also means that in this commit while policies are added the auth
   model is still password authorization based.
7. Next set of commits should span beyond the python scripts and implement
   calls to policy based authorization.

Signed-off-by: Imran Desai <imran.desai@intel.com>


src/lib/object.h
```diff
@@ -54,6 +54,8 @@ struct tobject {
     uint32_t handle;     /** loaded tpm handle */
 
     bool is_authenticated; /** true if a context specific login has authenticated use of the object */
+
+    CK_OBJECT_HANDLE policytype;
```
williamcroberts: why CK_OBJECT_HANDLE?

tools/tpm2_pkcs11/commandlets_keys.py
```diff
@@ -472,16 +477,32 @@ def generate_options(self, group_parser):
...
+        tertiary_object_policy_type = 0
+        policy = None
+        if nopolicy == False:
+            print("Adding USER OBJECT policy to CREATE")
+            tertiary_object_policy_type = 5
+            with Db(path) as db:
+                policy = "policy"
+                db.getpolicydigest(5, policy)
+        else:
+            print("WARN:: NO POLICY")
```
williamcroberts: Why?

tools/tpm2_pkcs11/commandlets_store.py
```diff
@@ -60,6 +65,57 @@ def __call__(self, parser, args, values, option_string=None):
...
 
+    @staticmethod
+    def generate_policies_and_update_store(db, tpm2):
+
+        if db.is_policy_store_initialized():
+            return
+        #
+        # Policy ID/TYPE = 0 for objects that don't use policy for authorization
+        #
+        db.addpolicy(0)
```
williamcroberts: probably create constant values for this POLICY_NONE, POLICY_A, POLICY_B etc

tools/tpm2_pkcs11/commandlets_keys.py
idesai: Debug - leftover -- will cleanup

idesai:
> This doesn't look terrible :-p

There is one change I can think it needs :) Should a policy store exist on a per token basis?

idesai:
> > This doesn't look terrible :-p
> 
> There is one change I can think it needs :) Should a policy store exist on a per token basis?

Reason being, something like policysecret would rely on a different set of SO/ USER PIN objects for every token.

williamcroberts:
Just closing this for now (zero inbox the PR queue), re-open when ready

PR #328, opened 2019-11-26
re-enable python 3 testing

williamcroberts:
Figured out the issue, for details see commit messages and https://github.com/etingof/pyasn1/issues/185


williamcroberts:
Yeah I go those today

PR #329, opened 2019-11-27
test: use relative path to determine location of libtpm2_pkcs11

diabonas:
Running the integration test outside Travis CI fails because `create_pkcs_store.sh` relies on a hardcoded `/workspace` build directory. Specify locations relative to the current build directory `$PWD` instead to make the tests portable.


williamcroberts:
Why did I make this so hard...

PR #339, opened 2019-12-04
Fix SPDX license identifiers

diabonas:
The correct identifier is "BSD-2-Clause", not "BSD-2", see https://spdx.org/licenses/BSD-2-Clause.html


williamcroberts:
Thanks, this was also on my todo list!

Issue #361, opened 2019-12-13
tpm2_ptool: init fails with py3 unless --primary-auth is given


dkobras:
Trying to run `tpm2_ptool init` with a Python 3 interpreter fails with the following error:
```
PYTHON_INTERPRETER=python3 tpm2_ptool init
Traceback (most recent call last):
  File "/src/tpm2/tpm2-pkcs11/tools/tpm2_pkcs11/commandlets_store.py", line 87, in __call__
    ctx = tpm2.createprimary(ownerauth, pobjauth)
  File "/src/tpm2/tpm2-pkcs11/tools/tpm2_pkcs11/tpm2.py", line 29, in createprimary
    stderr)
RuntimeError: Could not execute tpm2_createprimary: b'WARNING:esys:src/tss2-esys/api/Esys_CreatePrimary.c:389:Esys_CreatePrimary_Finish() Received TPM Error \nERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:135:Esys_CreatePrimary() Esys Finish ErrorCode (0x000001d5) \nERROR: Esys_CreatePrimary(0x1D5) - tpm:parameter(1):structure is the wrong size\nERROR: Unable to run tpm2_createprimary\n'
Could not execute tpm2_createprimary: b'WARNING:esys:src/tss2-esys/api/Esys_CreatePrimary.c:389:Esys_CreatePrimary_Finish() Received TPM Error \nERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:135:Esys_CreatePrimary() Esys Finish ErrorCode (0x000001d5) \nERROR: Esys_CreatePrimary(0x1D5) - tpm:parameter(1):structure is the wrong size\nERROR: Unable to run tpm2_createprimary\n'
```

That's because `binascii.hexlify()` returns a `bytes()` object in Python3. Via the `rand_hex_str()` function in `utils.c`, this leads to `tpm2_createprimary` getting called with an argument list like `["tpm2_createprimary", "-p", "b'9ff0e4a56086b8bdc88034f5943ef41e'", ...]`. Decoding the `bytes()` object to a string before passing it on, fixes the problem.

williamcroberts:
Fixed: https://github.com/tpm2-software/tpm2-pkcs11/pull/360

PR #362, opened 2019-12-13
tpm2_ptool: fix py3 error with init

dkobras:
Pass auto-generated primary authentication values as strings, so the
subsequent call to tpm2_createprimary receives the correct option value.
Otherwise, Python 3 would pass the string representation of a bytes()
object that tpm2_createprimary cannot handle.

Fixes: #361

Signed-off-by: Daniel Kobras <kobras@puzzle-itc.de>

### DIFF
--- a/tools/tpm2_pkcs11/commandlets_store.py
+++ b/tools/tpm2_pkcs11/commandlets_store.py
@@ -83,7 +83,7 @@ class InitCommand(Command):
 
                     if not use_existing_primary:
                         pobjauth = pobjauth if pobjauth != None else rand_hex_str(
-                        )
+                        ).decode('ascii')
                         ctx = tpm2.createprimary(ownerauth, pobjauth)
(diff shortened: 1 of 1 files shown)


tools/tpm2_pkcs11/commandlets_store.py
```diff
@@ -83,7 +83,7 @@ def __call__(self, args):
 
                     if not use_existing_primary:
                         pobjauth = pobjauth if pobjauth != None else rand_hex_str(
-                        )
+                        ).decode('ascii')
```
williamcroberts: I have a similar fix in: https://github.com/tpm2-software/tpm2-pkcs11/pull/360.

williamcroberts:
A slightly different way of doing it is In PR: https://github.com/tpm2-software/tpm2-pkcs11/pull/360

PR #366, opened 2019-12-17
support wpa supplicant

williamcroberts:
This substantially fixes the mechanism handling code and has a few other minor tweaks.


dnoliver:
Some minor comments:

1. bash scripts: some indentation details, mixing 2 and 4 spaces. I saw the [ShellCheck](https://www.shellcheck.net/) tool being used in other projects (latchset/clevis) to validate bash scripts. Maybe will be useful to add to travis. Also, maybe will help to `set -euxo pipefail` just in case.
2. Is there any concern on still using SHA1 with PKCS11? 

Thank you @williamcroberts!!

williamcroberts:
Looking into the failure has been nasty. For some reason when I run version 1.0.2g in the script vs manually in the container I am seeing issues around this line:
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2g/apps/s_server.c#L2362

On failure, both FD's are in the FDSET from select causing the read_from_terminal and read_from_sslcon to be set to 1. This causes the read from terminal path to try and read from stdin, which is empty. I can see this in the strace between working and non working attempting FD 4 vs FD 0 (stdin). Not sure what the best way to rectify this is yet.

williamcroberts:
A simple program demonstrates the behavior difference between running normally and in the background. It looks like stdin is still connected to the terminal...

```C
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>
#include <sys/select.h>

int main() {

 fd_set readfds;
 FD_ZERO(&readfds);
 FD_SET(0, &readfds);
 int rc = select(1, &readfds, NULL, NULL, NULL);
 printf("rc: %d\n", rc);
 printf("is fd 0 set: %s\n", FD_ISSET(0, &readfds) ? "yes" : "no");

 return rc;
}
```

williamcroberts:
Looks like I can mkfifo and then use that as stdin.

williamcroberts:
> Looks like I can mkfifo and then use that as stdin.

Nope it needs one byte for whatever reason... I'm not sure if this will be a check openssl version and then do thing or not

williamcroberts:
> Some minor comments:
> 
> 1. bash scripts: some indentation details, mixing 2 and 4 spaces. I saw the [ShellCheck](https://www.shellcheck.net/) tool being used in other projects (latchset/clevis) to validate bash scripts. Maybe will be useful to add to travis. Also, maybe will help to `set -euxo pipefail` just in case.

Good suggestions, ill have to look up what all those flags do.

> 2. Is there any concern on still using SHA1 with PKCS11?

From a crypto standpoint, yes. SHA1 is considered broken. But if you're using it for your hashtable and your hashtable can withstand collisions (like any good hashtable can), sha1 would be fine.
Our goal here is to provide mechanism, not policy. So if the TPM supports it, we expose it.

> 
> Thank you @williamcroberts!!

PR #378, opened 2020-01-02
C init token support

williamcroberts:
- Depends on CI build changes:
  - https://github.com/tpm2-software/tpm2-software-container/pull/22

Initial code for supporting C_InitToken, on 1 of three states:
1. db found with primary object configured, use that.
2. no db found, use PC WG Spec SRK if present at address 0x81000001
3. no db found, create a persistent key at the first persistent slot.

**This contains a schema bump and upgrades the DB**

**Backup your db**
I would recommend, since this is the first upgrade schema code, to manually backup your db.
Backup instructions can be found here:
  - https://stackoverflow.com/questions/25675314/how-to-backup-sqlite-database

### DIFF
--- a/src/lib/token.c
+++ b/src/lib/token.c
@@ -23,6 +23,33 @@
 #include "token.h"
 #include "utils.h"
 
+CK_RV token_min_init(token *t) {
+    /*
+     * Initialize the per-token session table
+     */
+    CK_RV rv = session_table_new(&t->s_table);
+    if (rv != CKR_OK) {
+        LOGE("Could not initialize session table");
+        return rv;
+    }
+
+    /*
+     * Initialize the per-token tpm context
+     */
+    rv = tpm_ctx_new(t->config.tcti, &t->tctx);
+    if (rv != CKR_OK) {
+        LOGE("Could not initialize tpm ctx: 0x%lx", rv);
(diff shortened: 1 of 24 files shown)


tools/setup.py
```diff
@@ -9,6 +9,7 @@
 
 setup(
     name='tpm2-pkcs11-tools',
+    python_requires='>3.8',
```
williamcroberts: >=3.7

src/lib/token.c
```diff
@@ -115,6 +142,132 @@ CK_RV token_get_info (token *t, CK_TOKEN_INFO *info) {
     return CKR_OK;
 }
 
+CK_RV get_or_create_primary(token *t) {
+
+    twist blob = NULL;
+
+    /* if their is no primary object ... */
```
williamcroberts: There

src/lib/token.c
```diff
@@ -115,6 +142,132 @@ CK_RV token_get_info (token *t, CK_TOKEN_INFO *info) {
...
+    if (t->pid) {
+        return CKR_OK;
+    }
+
+    /* is their one in the db to use ? */
```
williamcroberts: There

Issue #341, opened 2019-12-06
Could not create RSA public key


timdesi:
I am testing with Java 13, IBM TMP2 (1332) simulator and latest code from master brunch.

When try to use Cipher - RSA/ECB/NoPadding for Encryption PKCS11 provider try to use native C_CreateObject to generate public key and exit with error below. This function looks that is not implemented yet.

If you need further information or sample application let me know.

```

   java.security.InvalidKeyException: Could not create RSA public key
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11RSAKeyFactory.implTranslatePublicKey(P11RSAKeyFactory.java:71)
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11KeyFactory.engineTranslateKey(P11KeyFactory.java:127)
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11KeyFactory.convertKey(P11KeyFactory.java:64)
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11RSACipher.implInit(P11RSACipher.java:200)
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11RSACipher.engineInit(P11RSACipher.java:168)
	at java.base/javax.crypto.Cipher.init(Cipher.java:1299)
	at java.base/javax.crypto.Cipher.init(Cipher.java:1239)
	at javaapplication4.JavaApplication4.main(JavaApplication4.java:108)
  Caused by: sun.security.pkcs11.wrapper.PKCS11Exception: CKR_FUNCTION_NOT_SUPPORTED
	at jdk.crypto.cryptoki/sun.security.pkcs11.wrapper.PKCS11.C_CreateObject(Native Method)
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11RSAKeyFactory.generatePublic(P11RSAKeyFactory.java:198)
	at jdk.crypto.cryptoki/sun.security.pkcs11.P11RSAKeyFactory.implTranslatePublicKey(P11RSAKeyFactory.java:57)
```

williamcroberts:
Can you please provide sample code?

timdesi:
Sample code and console output is below;

```

package tpm2;

import java.security.Key;
import static org.junit.jupiter.api.Assertions.*;

import java.security.KeyStore;
import java.security.Provider;
import java.security.cert.X509Certificate;
import java.util.ArrayList;
import java.util.Collections;
import java.util.Enumeration;
[...]

PR #386, opened 2020-01-15
C_CreateObject: minimal RSA Java support

williamcroberts:
This creates basic support for RSA public key crypto for algorithm
RSA/ECB/RSAPKCS1.5 padding in Java. Other schemes may work, but are
untested. Test machines must have the hamcrest and junit JAR files on
their CLASSPATH.

Fixes: #341


williamcroberts:
re-build after docker image updates

PR #393, opened 2020-01-22
 PKCS11JavaTests: Java >= 9 compatibility

diabonas:
In Java >=9, the PKCS#11 provider needs to be configured as follows:
```
String pkcs11Config = "--name = TPM2\nlibrary = " + libPath;
PROV = Security.getProvider("SunPKCS11");
PROV = PROV.configure(pkcs11Config);
```
The `sun.security.pkcs11.SunPKCS11` class is not user-accessible any more, trying to instantiate it leads to a "package sun.security.pkcs11 is not visible" compilation error.

In Java <= 8, the `Provider.configure()` method does not exist, the configuration needs to be supplied to the class constructor instead.

To support all Java versions, check if `Provider.configure()` is available and fall back to the old code in case it is not. To avoid compile-time "symbol not found" errors due to missing/inaccessible methods and classes, we need to use introspection to do the resolution during runtime.

Additionally check during configure if a Java compiler can be found and whether the JUnit and Hamcrest classes are available.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -102,6 +102,13 @@ AC_ARG_ENABLE(
     [build and execute integration tests])],,
   [enable_integration=no])
 
+# Test for Java compiler and interpreter without throwing fatal errors (since
+# these macros are defined using AC_DEFUN they cannot be called conditionally)
+m4_pushdef([AC_MSG_ERROR], [have_javac=no])
+AX_PROG_JAVAC()
+AX_PROG_JAVA()
+m4_popdef([AC_MSG_ERROR])
+
 AC_DEFUN([integration_test_checks], [
 
   PKG_CHECK_MODULES([CMOCKA],[cmocka])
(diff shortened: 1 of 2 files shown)


diabonas:
Sadly the configure checks get called regardless of whether integration tests are enabled or not due to Autoconf peculiarities, so I am dropping 5c85b9113d6c8d2839ee08854b7db73ba83da058 for now.

diabonas:
Using some Autoconf trickery to temporarily make the checks for a working Java environment non-fatal, I was able to make the configure checks work as expected for both integration tests enabled and disabled.

configure.ac
```diff
@@ -102,6 +102,13 @@ AC_ARG_ENABLE(
     [build and execute integration tests])],,
   [enable_integration=no])
 
+# Test for Java compiler and interpreter without throwing fatal errors (since
+# these macros are defined using AC_DEFUN they cannot be called conditionally)
+m4_pushdef([AC_MSG_ERROR], [have_javac=no])
+AX_PROG_JAVAC()
+AX_PROG_JAVA()
+m4_popdef([AC_MSG_ERROR])
```
williamcroberts: I was looking for this stuff... thank you very much!

test/integration/PKCS11JavaTests.java
```diff
@@ -19,6 +19,11 @@
 import org.junit.runner.Result;
 import org.junit.runner.notification.Failure;
 
+/* Introspection to support Java < 9 */
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
```
williamcroberts: Gotta love reflection...

PR #402, opened 2020-02-04
Improve NSS support

dkobras:
Initialize required PKCS11 attributes to make NSS recognize certificates from `tpm2_pkcs11` by default. This is useful eg. for web browsers or S/MIME certificates in mail clients that are based on `libnss` and `libsmime`. Add a new integration test for the S/MIME use case.

### DIFF
--- a/test/integration/nss-tests.sh
+++ b/test/integration/nss-tests.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: BSD-2-Clause
+
+set -xo pipefail
+
+CA_PEM="$TEST_FIXTURES/ca.pem"
+CA_KEY="$TEST_FIXTURES/ca.key"
+
+CLIENT_CNF="$TEST_FIXTURES/client.cnf"
+PASSWORD_CA=whatever
+EXT_FILE="$TEST_FIXTURES/smimeextensions"
+
+export NSS_DEFAULT_DB_TYPE=sql
+
+if ! command -v certutil ||
+   ! command -v modutil; then
+  # Skip this test unless certutil/modutil are found
+  exit 77
+fi
+
+function pinentry() {
+  pin="$1"
+  shift
+  cmd="$*"
+  printf 'spawn %s\nexpect "Password or Pin"\nsend -- %s\\r\nexpect eof\n' \
+	  "$cmd" "$pin" | expect
+}
+
+function cleanup() {
+  if [ "$1" != "no-kill" ]; then
+      pkill -P $$ || true
+  fi
+  rm -f index.txt index.txt.attr serial serial.old index.txt.old index.txt.attr.old \
+        03.pem smimeclient.csr smimeclient.crt smimeclient.key smimeclient.pem \
+        pkcs11.txt cert9.db key4.db userpin.txt
+}
+trap cleanup EXIT
+
+onerror() {
+  echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
+  exit 1
+}
+trap onerror ERR
+
+cleanup "no-kill"
+
+if [ -z "$modpath" ]; then
+  modpath="$PWD/src/.libs/libtpm2_pkcs11.so"
+fi
+
+echo "modpath=$modpath"
+
+# make the DB
+touch index.txt
+touch index.txt.attr
+echo "03" >> serial
+
+echo "Creating S/MIME certificate for testuser@example.org"
+openssl req -new -newkey rsa:2048 -nodes -keyout smimeclient.key \
+  -out smimeclient.csr \
+  -subj "/C=FR/ST=Radius/L=Somewhere/O=Example Inc./CN=smimetesting/emailAddress=testuser@example.org"
+
+echo "Signing S/MIME certificate with test CA"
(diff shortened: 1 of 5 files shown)


dkobras:
Note that the NSS integration test is skipped by default because the default test image currently doesn't include the required package for `certutil` and `modutil`. So far, the configuration below `.ci/` is rather distribution agnostic, and I didn't want to change this with a hard-coded `apt-get install -y libnss3-tools`. Could it be added to the test image, instead?

williamcroberts:
> Note that the NSS integration test is skipped by default because the default test image currently doesn't include the required package for `certutil` and `modutil`. So far, the configuration below `.ci/` is rather distribution agnostic, and I didn't want to change this with a hard-coded `apt-get install -y libnss3-tools`. Could it be added to the test image, instead?

yes, ill add it to the Dockerfile.

test/integration/nss-tests.sh
```diff
@@ -0,0 +1,103 @@
...
+if [ -z "$modpath" ]; then
+  modpath="$PWD/src/.libs/libtpm2_pkcs11.so"
+fi
+
+echo "modpath=$modpath"
+
+# make the DB
+touch index.txt
+touch index.txt.attr
+echo "03" >> serial
```
williamcroberts: copy paste error? Why are you creating the db? Does something below use it?

test/integration/nss-tests.sh
```diff
@@ -0,0 +1,103 @@
...
+	  "$cmd" "$pin" | expect
+}
+
+function cleanup() {
+  if [ "$1" != "no-kill" ]; then
+      pkill -P $$ || true
+  fi
+  rm -f index.txt index.txt.attr serial serial.old index.txt.old index.txt.attr.old \
+        03.pem smimeclient.csr smimeclient.crt smimeclient.key smimeclient.pem \
+	pkcs11.txt cert9.db key4.db userpin.txt
```
williamcroberts: indent error

test/integration/nss-tests.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
+
+set -xo pipefail
+
+# SPDX-License-Identifier: BSD-2-Clause
```
williamcroberts: I know this is a copy + paste issue from my mistake, can you make this line no 2.

test/integration/nss-tests.sh
dkobras: It's required by `openssl ca` below. Usually, `tls-tests.sh` (where this part and much of the framework as a whole is copied from) will have it set up already, but this code makes sure `nss-tests.sh` can also be called stand-alone.

test/integration/nss-tests.sh
dkobras: Thanks, fixed!

test/integration/nss-tests.sh
dkobras: Fixed as well.

test/integration/nss-tests.sh
williamcroberts: Ahh I scanned for that command, I missed it.

williamcroberts:
This looks good, Ill try and get the relevant packages onto the dockerfile and this pulled in this week. Probably Friday.

test/integration/nss-tests.sh
```diff
@@ -0,0 +1,103 @@
...
+CLIENT_CNF="$TEST_FIXTURES/client.cnf"
+PASSWORD_CA=whatever
+EXT_FILE="$TEST_FIXTURES/smimeextensions"
+
+export NSS_DEFAULT_DB_TYPE=sql
+
+if ! command -v certutil ||
+   ! command -v modutil; then
+  # Skip this test unless certutil/modutil are found
+  exit 77
```
diabonas: Other tests like [`pkcs11-tool.sh`](https://github.com/tpm2-software/tpm2-pkcs11/blob/master/test/integration/pkcs11-tool.sh) just give a hard failure if the required tools are not found. I think this is preferable since it guarantees that tests are not accidentally skipped due to a misconfigured build environment.

I suggest making this check return a hard error (`exit 99`) and adding an `AC_CHECK_PROG` for `certutil` and `modutil` to `configure.ac`: https://github.com/tpm2-software/tpm2-pkcs11/blob/2dbe1ffc582c9aa8521317e1e9eb5daa39ef4c80/configure.ac#L112-L132 Additionally these tools should be included in the [`INSTALL.md`](https://github.com/tpm2-software/tpm2-pkcs11/blob/master/docs/INSTALL.md#executable-dependencies) instructions.

williamcroberts: @diabonas he actually points this out in his message that its a TODO item, I plan on fixing it when I pull it in. I plan on working on this next week, im a bit behind.

williamcroberts:
Update: I'll be looking at this next week. I was under the weather, sorry for the delays.

williamcroberts:
Dockerfile updated:
https://github.com/tpm2-software/tpm2-software-container/commit/00297d5dda1ac6cb17e6ecb4369cb83a401dde03

williamcroberts:
Moved with conflict fixed and patch updated to check for cert and modutil in configure. Waiting for Docker image to update and will have to retrigger tests

https://github.com/tpm2-software/tpm2-pkcs11/pull/423

PR #414, opened 2020-02-21
FAPI backend

AndreasFuchsSIT:
The current status is as follows:
- SO-seal is created in FAPI-store
- USR-seal is created in FAPI-store
- tobjects are part of so-seal appdata
- tokens of SQLite and FAPI are concatenated (so both work in parallel)
- test pkcs11-tool-init{-fapi} uses FAPI

TODOs:
- emptytoken comes from db.c instead of token.c
- Somewhat hacky concatenation (dep: emptytoken)
- pobjects are a bit hacky (need to be made Flushable for FAPI)
- other tests need to learn about FAPI; tpm2_ptool or something else ?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,11 +5,11 @@ ACLOCAL_AMFLAGS = -I m4 --install
 AM_CFLAGS       = $(INCLUDE_DIRS) $(EXTRA_CFLAGS) $(CODE_COVERAGE_CFLAGS) \
                   $(TSS2_ESYS_CFLAGS) $(TSS2_MU_CFLAGS) $(TSS2_TCTILDR_CFLAGS) \
 		  $(TSS2_RC_CFLAGS) $(SQLITE3_CFLAGS) $(PTHREAD_CFLAGS) \
-		  $(CRYPTO_CFLAGS) $(YAML_CFLAGS)
+		  $(CRYPTO_CFLAGS) $(YAML_CFLAGS) $(TSS2_FAPI_CFLAGS)
 
 AM_LDFLAGS      = $(EXTRA_LDFLAGS) $(CODE_COVERAGE_LIBS) $(TSS2_ESYS_LIBS) \
                   $(TSS2_MU_LIBS) $(TSS2_TCTILDR_LIBS) $(TSS2_RC_LIBS) \
-		  $(SQLITE3_LIBS) $(PTHREAD_LIBS) $(CRYPTO_LIBS) $(YAML_LIBS)
+		  $(SQLITE3_LIBS) $(PTHREAD_LIBS) $(CRYPTO_LIBS) $(YAML_LIBS) $(TSS2_FAPI_LIBS)
 
 check-programs: $(check_PROGRAMS)
 
@@ -133,6 +133,15 @@ check_PROGRAMS += \
 
 # add test scripts
 check_SCRIPTS += $(integration_scripts)
+
+#Cannot be part of $(integration_scripts) since its not EXTRA_DIST
+check_SCRIPTS += test/integration/pkcs11-tool-init-fapi.sh.nosetup
+
+test/integration/pkcs11-tool-init-fapi.sh.nosetup: test/integration/pkcs11-tool-init.sh.nosetup
+	$(AM_V_GEN)(echo "#!/usr/bin/env bash"; echo "export FAPI_PREVIEW=true"; cat $^)>$@ \
+		&& chmod +x "$@"
+
+CLEANFILES += test/integration/pkcs11-tool-init-fapi.sh.nosetup
 
 LOG_COMPILER = $(srcdir)/test/integration/scripts/int-test-setup.sh
 
--- a/src/lib/backend.c
+++ b/src/lib/backend.c
@@ -1,0 +1,50 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include "config.h"
+#include "backend.h"
+#include "backend_esysdb.h"
+
+/* This file includes the logic for selecting, aggregating and
+ * distributing calls to different backends.
+ * For now this will only be the esysdb backend that uses tss2-esys
+ * and sqlite3 for operations.
+ * In the future, logic will be added to also inlcude the tss2-fapi
+ * library for storage and TPM interaction.
+ */
+
+CK_RV backend_init(void) {
+    return backend_esysdb_init();
+}
+
+CK_RV backend_destroy(void) {
+    return backend_esysdb_destroy();
+}
+
+/** Create a new token
+ *
+ * Create a new sealed object and store it in the data store.
+ *
+ * @param[in,out] t The token information on input and generated token
+ *                  on output.
+ * @param[in] hexwrappingkey TODO
+ * @param[in] newauth The authorization value for the security operator
+ *                    of the newly created token.
+ * @param[in] newsalthex TODO
+ * @returns TODO
+ */
+CK_RV backend_create_token_seal(token *t, const twist hexwrappingkey,
+                       const twist newauth, const twist newsalthex) {
+    return backend_esysdb_create_token_seal(t, hexwrappingkey, newauth, newsalthex);
+}
+
+/** Retrieve the all tokens available.
+ *
+ * The returned list is a set of all stored tokens with all
+ * objects inside the token structure.
+ * @param[out] tok The list of tokens.
+ * @param[out] len The number of entries in tok.
+ * @returns TODO
+ */
+CK_RV backend_get_tokens(token **tok, size_t *len) {
+    return backend_esysdb_get_tokens(tok, len);
+}
--- a/src/lib/backend.c
+++ b/src/lib/backend.c
@@ -3,6 +3,7 @@
 #include "config.h"
 #include "backend.h"
 #include "backend_esysdb.h"
+#include "backend_fapi.h"
 
 /* This file includes the logic for selecting, aggregating and
  * distributing calls to different backends.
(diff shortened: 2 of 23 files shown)


test/integration/scripts/int-test-setup.sh
```diff
@@ -175,6 +175,9 @@ echo ${TPM2TOOLS_TCTI}
 export TPM2_PKCS11_TCTI="tabrmd:${TABRMD_TEST_TCTI_CONF}"
 echo ${TPM2_PKCS11_TCTI}
 
+setup_fapi
+tss2_provision
```
williamcroberts: As im sure your already aware, but ill just mention it again, you'll also need tests at some point where the setup isn't already done.

AndreasFuchsSIT:
Agreed. Basically I'll be separating out object creation and changeauth and get rid of the db in this first iteration.

williamcroberts:
> Agreed. Basically I'll be separating out object creation and changeauth and get rid of the db in this first iteration.

What exactly do you mean? We still need to be able to read from existing dbs or have a migration path forward for them.

AndreasFuchsSIT:
What I meant was: Provide an implementation using FAPI creation and keystore but keep current crypto handling via esys.
I will keep the db-based reading of existing tokens as is (besides some refactoring).

williamcroberts:
> What I meant was: Provide an implementation using FAPI creation and keystore but keep current crypto handling via esys.
> I will keep the db-based reading of existing tokens as is (besides some refactoring).

That's what I thought, but I was just wanted to be pedantic and make sure were still on the same page.

AndreasFuchsSIT:
@williamcroberts Would you mind having a look at this ?
IMHO, this could be the first merge into master, pending some more tests.

Which do you see as most important to get fapi-versions for ?

Makefile.am
```diff
@@ -134,6 +134,15 @@ check_PROGRAMS += \
...
 check_SCRIPTS += $(integration_scripts)
 
+#Cannot be part of $(integration_scripts) since its not EXTRA_DIST
+check_SCRIPTS += test/integration/pkcs11-tool-init-fapi.sh.nosetup
+
+test/integration/pkcs11-tool-init-fapi.sh.nosetup: test/integration/pkcs11-tool-init.sh.nosetup
+	$(AM_V_GEN)(echo "#!/usr/bin/env bash"; echo "export FAPI_PREVIEW=true"; cat $^)>$@ \
+		&& chmod +x "$@"
+
+CLEANFILES += test/integration/pkcs11-tool-init-fapi.sh.nosetup
```
williamcroberts: Why are we doing all this? Why wouldn't this just be like any other integration test?

src/lib/backend.c
```diff
@@ -0,0 +1,151 @@
...
+ * For now this will only be the esysdb backend that uses tss2-esys
+ * and sqlite3 for operations.
+ * In the future, logic will be added to also inlcude the tss2-fapi
+ * library for storage and TPM interaction.
+ */
+
+CK_RV backend_init(void) {
+    LOGV("Initializing backends");
+    CK_RV rv = backend_fapi_init();
+    if (rv)
```
williamcroberts: brackets on everything.

src/lib/backend.c
```diff
@@ -0,0 +1,151 @@
...
+/** Retrieve all tokens available.
+ *
+ * The returned list is a set of all stored tokens with all
+ * objects inside the token structure.
+ * @param[out] tok The list of tokens.
+ * @param[out] len The number of entries in tok.
+ * @returns TODO
+ */
+CK_RV backend_get_tokens(token **tok, size_t *len) {
+    CK_RV rv;
```
williamcroberts: declare and assign on line 84.

src/lib/backend.c
```diff
@@ -0,0 +1,151 @@
...
+
+    /* This is used to move the empty token to the end. */
+    /* TODO: Would be better to have a nicer way of doing so. */
+    token tmp;
```
williamcroberts: assign and declare on 91.

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include "config.h"
+#include "backend_fapi.h"
+#include "emitter.h"
+#include "parser.h"
+#include <tss2/tss2_fapi.h>
+
+FAPI_CONTEXT *fctx = NULL;
+int objectids = 0;
```
williamcroberts: are these meant to be extern? You probably want static.

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+    return CKR_OK;
+}
+
+CK_RV backend_fapi_ctx_new(token *t) {
+    t->fapi.ctx = fctx;
+    return CKR_OK;
+}
+
+void backend_fapi_ctx_free(token *t) {
+    (void)(t);
```
williamcroberts: #include "util.h"
UNUSED(t);

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+}
+
+
+#define PREFIX "/HS/SRK/tpm2-pkcs11-token-"
+
+static char * tss_path_from_id(unsigned id, const char *type) {
```
williamcroberts: do you modify these afterwards? const char * seems better.

Also, * tight to the variable like what you're doing on line 45.

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+    char *path = malloc(strlen(PREFIX) + strlen(type) + 1 + 8 + 1);
```
williamcroberts: if you're doing math with magic numbers, explain why 1, 1 and 8.

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+ * See backend_create_token_seal()
+ */
+CK_RV backend_fapi_create_token_seal(token *t, const twist hexwrappingkey,
+                       const twist newauth, const twist newsalthex) {
+    TSS2_RC rc;
+    uint8_t *tpm2bPublic, *tpm2bPrivate, *appdata;
+    size_t tpm2bPublicSize, tpm2bPrivateSize, appdata_len;
+    twist pub, priv;
+
+    char *path = tss_path_from_id(t->id, "so");
```
williamcroberts: this can return NULL, is Fapi_CreateSeal fail with NULL path?

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+        /* Initialize the User PIN area. */
+        /*********************************/
+        path = tss_path_from_id(t->id, "usr");
+        if (!path) {
+            LOGE("OOM");
+            goto error;
+        }
+        rc = Fapi_GetTpmBlobs(t->fapi.ctx, path, &tpm2bPublic, &tpm2bPublicSize,
+                              &tpm2bPrivate, &tpm2bPrivateSize, NULL /* policy */);
+        if (rc == 0x00060020) {
```
williamcroberts: That's a very specific RC, is their any RC defines you can OR in to make it more clear?

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+
+        rc = Fapi_GetAppData(t->fapi.ctx, path, &appdata, NULL);
+        if (rc) {
+            LOGE("Getting FAPI seal appdata failed.");
+            goto error;
+        }
+
+        t->sealobject.userauthsalt = twistbin_new(appdata, strlen((char *)appdata));
+        Fapi_Free(appdata);
+        if (t->sealobject.userauthsalt == NULL) {
```
williamcroberts: previous style was to check for !x, this one does x == NULL, consistency.

src/lib/backend_fapi.c
```diff
@@ -0,0 +1,490 @@
...
+/** Create a new token in fapi backend.
+ *
+ * See backend_create_token_seal()
+ */
+CK_RV backend_fapi_create_token_seal(token *t, const twist hexwrappingkey,
+                       const twist newauth, const twist newsalthex) {
+    TSS2_RC rc;
+    uint8_t *tpm2bPublic, *tpm2bPrivate, *appdata;
+    size_t tpm2bPublicSize, tpm2bPrivateSize, appdata_len;
+    twist pub, priv;
```
williamcroberts: Generally speaking, unless their is goto scope issues, this projects coding style is to declare variables nearest to use and also not to do the list of types. One variable declaration per line. This applies here and throughout.

src/lib/token.h
```diff
@@ -50,20 +50,39 @@ struct sealobject {
...
     unsigned id;
     unsigned pid;
     unsigned char label[32];
 
-    token_config config;
+    enum token_type type;
 
-    pobject pobject;
+//TODO: uncomment union, once both backends are completely separated.
+//    union { /* anon union for backend data */
```
williamcroberts: I thought we we're merging the backends, so you can have N ESYS DB tokens + M FAPI tokens?

test/integration/pkcs11-tool-init.sh.nosetup
```diff
@@ -25,6 +27,7 @@ fi
 echo "modpath=$modpath"
 
 pkcs11_tool() {
+#  valgrind pkcs11-tool --module "$modpath" "$@"
```
williamcroberts: delete me.

src/lib/backend.c
williamcroberts: Perhaps don't let the backends create the empty token, move it into the backend.c code.

Makefile.am
AndreasFuchsSIT: Well, I wanted to ensure 1-to-1 compatibility between FAPI and non-FAPI and as such I wanted to run the exact same test with FAPI and with non-FAPI.
I guess I want to try the TEST_ENVIRONMENT instead. But that would still need a symlink from .fapi to .nosetup. I guess that's nicer...

src/lib/backend_fapi.c
AndreasFuchsSIT: @williamcroberts I don't think I understand this comment ? are you referring to the return value ?

AndreasFuchsSIT: Because I don't want to prevent modification of the returned value.

Makefile.am
AndreasFuchsSIT: Ok, so there is no possibility of `FAPI_TEST_ENVIRONMENT`, only a `FAPI_LOG_COMPILER` that I could use to make this work...

PR #419, opened 2020-02-24
build: always distribute test files

diabonas:
The release tarball for the current [release candiate](https://github.com/tpm2-software/tpm2-pkcs11/releases/tag/1.1.0-RC0) appears to have been build without `--enable-integration`. This means the integration test suite cannot be run because some required scripts are distributed depending on the state of `ENABLE_INTEGRATION`.  Make sure that `EXTRA_DIST` is independent of the configuration options to avoid such problems in the future. Additionally there are two leftover files that `make distcheck` may complain about, make sure to remove these.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,6 +91,19 @@ check_SCRIPTS =
 
 TEST_EXTENSIONS = .sh .int .nosetup .java
 
+integration_scripts = \
+    test/integration/pkcs11-tool.sh \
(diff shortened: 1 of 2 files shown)


williamcroberts:
on 1.1.X as 82e2dfa29b5ea1b7ef5bd3d9cda1dda8ba380af9 and 8443794bd6d51071dc75a8dbe58f5d0adaaf9c6a

PR #421, opened 2020-02-25
test/pkcs11-dbup: use release tarball instead of git clone

diabonas:
When running the test suite from a release tarball, using a git clone requires autoconf-archive to be installed for the Autconf bootstrap process. Replacing this with an already bootstrapped release tarball for version 1.0 changes the required packages from git/autoconf-archive to wget/tar which are more widely available. Additionally pinning the checksum of the tarball reduces the risk of downloading and running code from the Internet.


williamcroberts:
on 1.1.X as 3a586a8d063defb9dbb6061c293f9f43e8df5984

PR #427, opened 2020-03-02
Replace PKCS11 OASIS headers with a version from p11-kit

alvinchchen:
Hello,

I am currently working on debian packaging tpm2-pkcs11.
And found headers provided by OASIS PKCS11 TC are not-exactly free license
(they do not allow modification), 
I use an alternative header from p11-kit
which is licensed under a more liberal license.

Regards,
Alvin


williamcroberts:
I guess we'll just switch than, because it seems to be confusing everyone, even though the original headers are fine and don't need any modifications.

Issue #432, opened 2020-03-04
tcti and log level config broken


williamcroberts:
Currently, tpm2_ptool allows for per token configuration of loglevel and tcti. However, these are really per-store variables. So modify:
1. bump the DB schema to support a config table that contains log-level and tcti string.
  - Default the value to the value stored in the first token
2. config command to modify this store-wide table, rather than per token config.

**This will be a non-backwards compat tooling change, so this will cause a major version bump.**

williamcroberts:
Per token TCTI contexts work, its just log level that needs go away.

PR #433, opened 2020-03-05
config: move per token config tcti/log-level

williamcroberts:
WARNING THIS PATCH BUMPS THE DB VERSION FROM 3 TO 4. AS ALWAYS BACKUP
YOUR DB UNDER THESE CONDITIONS.

WARNING THIS IS NON BACKWARDS COMPATIBLE CHANGES.

Currently, one can configure per-token configuration variables via the
tpm2_ptool config commandlet. The config variables "tcti" and
"log-level" were fundamentally broken by design. Each of those variables
is a process wide control, thus if tokens have conflicting options
between them, chaos will ensue.

To correct this, modify tpm2_ptool config commandlet to work on a store
wide config table. This is a non-backwards compatible change to
tpm2_ptool config commandlet, but should cause only minor issues as the
only commandline change is that --label is no longer required.

Bump the DB version to 4, and handle the upgrade paths.

Fixes: #432

### DIFF
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -38,7 +38,7 @@
 #define TPM2_PKCS11_STORE_DIR "/etc/tpm2_pkcs11"
 #endif
 
-#define DB_VERSION 3
+#define DB_VERSION 4
 
 #define goto_oom(x, l) if (!x) { LOGE("oom"); goto l; }
 #define goto_error(x, l) if (x) { goto l; }
@@ -386,7 +386,7 @@ CK_RV db_get_tokens(token **tok, size_t *len) {
                     LOGE("Expected token config to contain config data");
                     goto error;
                 }
-                bool result = parse_token_config_from_string(config, bytes, &t->config);
+                bool result = parse_token_config_from_string_v2(config, bytes, &t->config);
                 if (!result) {
                     LOGE("Could not parse token config, got: \"%s\"", config);
                     goto error;
@@ -806,7 +806,7 @@ CK_RV db_add_token(token *tok) {
 
     sqlite3_stmt *stmt = NULL;
 
-    char *config = emit_config_to_string(tok);
+    char *config = emit_token_config_to_string_v2(&tok->config);
     if (!config) {
         LOGE("Could not get token config");
         return CKR_GENERAL_ERROR;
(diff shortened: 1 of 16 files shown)


src/lib/db.c
```diff
@@ -1204,6 +1204,47 @@ static CK_RV db_get_version(sqlite3 *db, unsigned *version) {
...
+        const unsigned char *conf_yaml = sqlite3_column_text(stmt, 0);
+        size_t len = sqlite3_column_bytes(stmt, 0);
+        bool result = parse_store_config_from_string(
+                conf_yaml, len, config);
+        if (!result) {
+            LOGE("Could not parse store config, got: \"%s\"",
+                    conf_yaml);
+            goto error;
+        }
+
```
williamcroberts: empty, rm

src/lib/db.c
```diff
@@ -1204,6 +1204,47 @@ static CK_RV db_get_version(sqlite3 *db, unsigned *version) {
...
+    } else if (rc != SQLITE_DONE) {
+        /* something else happened (error), that wasn't empty db */
+        LOGE("Cannot step query: %s\n", sqlite3_errmsg(global.db));
```
williamcroberts: grong db reference in errmsg()

src/lib/db.c
```diff
@@ -1321,6 +1362,109 @@ static CK_RV dbup_handler_from_2_to_3(sqlite3 *updb) {
...
+
+    const unsigned char *yaml_config = NULL;
+    size_t yaml_config_len = 0;
+
+    rc = sqlite3_step(stmt);
+    if (rc == SQLITE_ROW) {
+        yaml_config = sqlite3_column_text(stmt, 0);
+        yaml_config_len = sqlite3_column_bytes(stmt, 0);
+    } else if (rc == SQLITE_DONE) {
+        goto out;
```
williamcroberts: this isn't an error and will trigger an error since RV is not set. goto done;

src/lib/db.c
```diff
@@ -1321,6 +1362,109 @@ static CK_RV dbup_handler_from_2_to_3(sqlite3 *updb) {
...
+    } else {
+        LOGE("Cannot step query: %s\n", sqlite3_errmsg(global.db));
+        goto out;
+    }
+
+    /* yaml config has something, deal with it */
+    if (yaml_config_len && (yaml_config || strcmp((const char *)yaml_config, ""))) {
```
williamcroberts: invert the logic and goto done to remove nesting level

src/lib/db.c
```diff
@@ -1321,6 +1362,109 @@ static CK_RV dbup_handler_from_2_to_3(sqlite3 *updb) {
...
+
+        /* Step 3: Set the new store data */
+        s = "REPLACE INTO config (id, config) VALUES (1, ?);";
+
+        sqlite3_stmt *stmt2;
+        int rc = sqlite3_prepare_v2(updb, s, -1, &stmt2, NULL);
+        if (rc != SQLITE_OK) {
+            LOGE("Cannot prepare config insert query: %s\n",
+                    sqlite3_errmsg(updb));
+            return rc;
```
williamcroberts: goto out

src/lib/parser.c
```diff
@@ -453,3 +540,156 @@ bool parse_token_config_from_string(const unsigned char *yaml, size_t size, toke
...
+    }
+
+    const char *value = NULL;
+    result = kvp_get_value("token-init", kvp_list, ARRAY_LEN(kvp_list), &value);
+    if (!result) {
+        LOGE("Could not retrieve value for key \"token-init\"");
+        goto out;
+    }
+
+    config->is_initialized = !strcmp(value, "true")
```
williamcroberts: if value is NULL error

williamcroberts: like done in V1

williamcroberts:
Nevermind, per-token TCTI contexts work.

Issue #399, opened 2020-02-04
pkcs11-tool: Deleting the public key after deleting the private key causes segfault


cplappert:
The deletion of the public key causes a segfault (l.11) if the private key was deleted before.
Only deleting the private key is not enough the delete the object (l.10).

```
01: export mod_path=PATH_TO_LIBTPM2_PKCS11.so
02: export TPM2_PKCS11_STORE="$HOME/store"
03: alias pkcs11_tool='pkcs11-tool --module $mod_path'
04: mkdir $TPM2_PKCS11_STORE

05: pkcs11_tool --init-token --slot=1 --label=myfirsttoken --so-pin=myfirstsopin
06: pkcs11_tool --init-pin --slot=1 --login --login-type="so" --so-pin=myfirstsopin --pin=myfirstuserpin
07: pkcs11_tool --keypairgen --slot=1 --label="mysignkey" --login --pin=myfirstuserpin 
08: pkcs11_tool --list-objects --slot=1

09: pkcs11_tool --delete-object --slot=1 --label="mysignkey" --pin=myfirstuserpin --login --type=privkey
10: pkcs11_tool --list-objects --slot=1

11: pkcs11_tool --delete-object --slot=1 --label="mysignkey" --pin=myfirstuserpin --login --type=pubkey
12: pkcs11_tool --list-objects --slot=1
```

williamcroberts:
Interesting, this repro's on the CI, but not locally. @cplappert the CI is using OpenSC package version 0.15.0 (as packaged in Ubuntu 16.04). What version are you using?

Looking at the stack trace, its nonsense, looks like something is really going wrong. Can you try
running against OpenSC version 0.17?

```bash
git clone https://github.com/OpenSC/OpenSC.git
git checkout 0.17.0
./bootstrap
# for debug symbols in so
./configure --disable-optimization 
make -j4
```
use the command for your testing:
OpenSC/src/tools/.libs/pkcs11-tool

CI Output w/ASAN:
```
==13448==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f5ae32071c4 bp 0x7ffdf9130260 sp 0x7ffdf9130210 T0)
==13448==The signal is caused by a READ memory access.
==13448==Hint: address points to the zero page.
    #0 0x7f5ae32071c3 in session_ctx_logout /workspace/tpm2-pkcs11/build/../src/lib/session_ctx.c:374:24
[...]

PR #435, opened 2020-03-05
test: add regression test for delete objects

williamcroberts:
This adds a regression test that deleting objects via pkcs11-tool,
specifically the private RSA object followed by the Public RSA object,
works.

Fixes: #399

### DIFF
--- a/src/lib/token.c
+++ b/src/lib/token.c
@@ -178,8 +178,14 @@ void token_rm_tobject(token *tok, tobject *t) {
     assert(tok->tobjects.head);
     assert(tok->tobjects.tail);
 
-    if (t == tok->tobjects.head) {
-        tok->tobjects.head = list_entry(tok->tobjects.head->l.next, tobject, l);
+    /* only item in the list */
+    if (t == tok->tobjects.tail &&
+            t == tok->tobjects.head) {
+        /* just empty the list */
+        tok->tobjects.head = tok->tobjects.tail = NULL;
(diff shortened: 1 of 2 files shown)


williamcroberts:
Once we have docker ubuntu 18.04 running, we can at least conditionally run this test with `DOCKER_TAG != ubuntu-16.04`. However, folks running local testing on UBUNTU 18.04 it might not be set, so we probably want to set it.

williamcroberts:
Interesting, the tests didn't indicate what I was expecting. With ASAN enabled, it fails on Ubuntu 16.04 and Ubuntu 18.04. I better check my list free code, again.

https://travis-ci.org/github/tpm2-software/tpm2-pkcs11/builds/661107860?utm_source=github_status&utm_medium=notification

williamcroberts:
What's even more odd, is that the keygen test has this code in it too:
```C
    /* delete that objects */
    rv = C_DestroyObject(session, pubkey);
    assert_int_equal(rv, CKR_OK);

    rv = C_DestroyObject(session, privkey); // Heap Use After Free
    assert_int_equal(rv, CKR_OK);
```

And none of that is failing or triggering any ASAN issues.

williamcroberts:
I think I see the problem:
```C
        tok->tobjects.head = list_entry(tok->tobjects.head->l.next, tobject, l);
```

in the case of the last object, the next pointer is null, but the code returns some wild pointer by accessing NULL - an offset, and sets head equal to it.

PR #440, opened 2020-03-10
Ubuntu 18.04

williamcroberts:


williamcroberts:
missing pkcs11-tool in opensc package, so adding to container. Will trigger build when ubuntu-18.04 container updates.

PR #443, opened 2020-03-12
Test: Extend pkcs11-tool tests

AndreasFuchsSIT:
Add tests for listing and reading objects as well as signing (verified using openssl).

This should trigger a bug in CI that I cannot reproduce locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
 # only one build should report coverage stats
 matrix:
   include:
-  - env: ENABLE_COVERAGE=true DOCKER_TAG=ubuntu-16.04
+  - env: ENABLE_COVERAGE=true DOCKER_TAG=ubuntu-18.04
     compiler: gcc
 
 script:
--- a/test/integration/pkcs11-tool-init.sh.nosetup
+++ b/test/integration/pkcs11-tool-init.sh.nosetup
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: BSD-2-Clause
 
+set -x
+
 source "$T/test/integration/scripts/helpers.sh"
 
 tempdir=$(mktemp -d)
@@ -25,6 +27,7 @@ fi
 echo "modpath=$modpath"
 
 pkcs11_tool() {
+#  valgrind pkcs11-tool --module "$modpath" "$@"
   pkcs11-tool --module "$modpath" "$@"
   return $?
 }
@@ -33,13 +36,20 @@ export TPM2_PKCS11_STORE="$tempdir"
 
 echo "TPM2_PKCS11_STORE=$TPM2_PKCS11_STORE"
 
+echo "testdata">${tempdir}/data
+
 echo "Should have unitialized token"
 pkcs11_tool -T | grep -q uninitialized
 echo "Found unitialized token"
 
+pkcs11_tool -I
+pkcs11_tool -T
+
 echo "Initializing token"
 pkcs11_tool --slot=1 --init-token --label=mynewtoken --so-pin=mynewsopin
 echo "Token initialized"
+
+pkcs11_tool -T
 
 echo "Initializing user pin"
 pkcs11_tool --slot=1 --login --login-type="so" --init-pin --so-pin=mynewsopin --pin=mynewuserpin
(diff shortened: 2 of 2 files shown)


AndreasFuchsSIT:
@williamcroberts Do you have any idea why this fails on Ubuntu 16.04 but runs on Ubuntu 18.04 ?

williamcroberts:
> @williamcroberts Do you have any idea why this fails on Ubuntu 16.04 but runs on Ubuntu 18.04 ?

```
+ pkcs11-tool --module /workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so -v --read-object --slot=1 --label=myrsakey --type pubkey --output-file /tmp/tmp.OzNVZKI9nU/pubkey.der
Using slot with ID 0x1
warning: PKCS11 function C_GetAttributeValue(VALUE) failed: rv = CKR_ATTRIBUTE_TYPE_INVALID (0x12)
error: get CKA_VALUE failed
```

I cant say for sure, but I've had issue with options between versions and pkcs11-tool and them not working properly. It appears that command is ignoring type and grabbing the private object, which will not have CKA_VALUE.

williamcroberts:
The only suggestion I have is to build the version ubuntu packaged and debug it and see precisely what its doing. You could also set TPM2_PKCS11_LOG_LEVEL=2 and see what its calling...

I know a lot of the options to would call and then filter.

AndreasFuchsSIT:
That would sound like an upstream bug that would allow me do skip the test on old versions of pkcs11-tool ?

williamcroberts:
> That would sound like an upstream bug that would allow me do skip the test on old versions of pkcs11-tool ?

So I can't replicate this on my Ubuntu 16.04 machine, and I even tried checking out 0.15 of OpenSC.

The fun part is figuring out what version of pkcs11-tool is running, theirs no dump version.

AndreasFuchsSIT:
That's nasty...

Does pkcs11-tool -I help ?
```
+ pkcs11_tool -I
+ pkcs11-tool --module /workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so -I
Using slot 0 with a present token (0x1)
Cryptoki version 2.40
```

williamcroberts:
> ```
> Cryptoki version 2.40
> ```

No thats the PKCS11 interface version. PKCS11 tool has been a thorn in my side. Maybe add your test to p11tool? Is their anything broken specifically you're looking at?

You could use pkgconfig to get the opensc version from configure and pass that along in the test ENV. The opensc package appears to always contain the pkcs11-tool + the library. So by checking the library version, you can implicitly figure out the pkcs11-tool version.

AndreasFuchsSIT:
Ok, willdo...

williamcroberts:
> So I can't replicate this on my Ubuntu 16.04 machine, and I even tried checking out 0.15 of OpenSC.
> 

I'm an idiot, I tested on Ubuntu 18.04.... So yeah, just get the opensc version and conditional on that.

AndreasFuchsSIT:
So I looked on my Ubuntu 18.04 that I have here and it does not come with a opensc.pc file.
Do you have one and I'm just stupid ?

I further tried looking at `opensc-tool --version` but that just says `No Git revision info available`.

Any ideas ?

williamcroberts:
> So I looked on my Ubuntu 18.04 that I have here and it does not come with a opensc.pc file.
> Do you have one and I'm just stupid ?
> 
> I further tried looking at `opensc-tool --version` but that just says `No Git revision info available`.
> 
> Any ideas ?

pkg-config --modversion opensc-pkcs11
0.19.0

AndreasFuchsSIT:
So apparently this .pc file is not packaged in Ubuntu at all:
```
$ apt-file search opensc-pkcs11.pc
```
Gives no results...

I've updated the test with a `|| exit 77` for the readpublic and moved it to the end of the file. That's as good as it gets on Ubuntu at least...

Also I've changed to code-cov docker to 18.04 so we report the bigger test coverage runner.

williamcroberts:
> So apparently this .pc file is not packaged in Ubuntu at all:
> 
> ```
> $ pkg-config --modversion opensc-pkcs11
> Package opensc-pkcs11 was not found in the pkg-config search path.
> 
> $ apt-file search opensc-pkcs11.pc || echo fail
> fail
> ```
> 
> I've updated the test with a `|| exit 77` for the readpublic and moved it to the end of the file. That's as good as it gets on Ubuntu at least...
> 
> Also I've changed to code-cov docker to 18.04 so we report the bigger test coverage runner.

Its packaged on Ubuntu-18.04+ but not 16.04. So if it's not there, just skip.

PR #446, opened 2020-03-12
configure: check for python >= 3.7

williamcroberts:
Use autoconf to check for Python version >= 3.7 when enabling
integration tests and automatically pass that version into the
PYTHON_INTERPRETER variable in the automake check ENV.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -39,9 +39,8 @@ if [[ "$CC" == clang* ]]; then
     echo "No llvm symbolizer found at: $ASAN_SYMBOLIZER_PATH"
     unset ASAN_SYMBOLIZER_PATH
   fi
-else #GCC
-  export ENABLE_COVERAGE=true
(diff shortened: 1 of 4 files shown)


williamcroberts:
Look like ubuntu16.04 has the same issue I noticed trying to get opensuse-leap working. The python.m4 file in automake, is essentially broken:
```m4
 m4_define_default([_AM_PYTHON_INTERPRETER_LIST],
[python python2 python3 python3.3 python3.2 python3.1 python3.0 python2.7 dnl
 python2.6 python2.5 python2.4 python2.3 python2.2 python2.1 python2.0])
```
Note that if it's not in the list, it fails.... automake 1.16 has it fixed....

williamcroberts:
For ubuntu, we can probably just avoid package automake, and install it manually:
http://mirrors.kernel.org/ubuntu/pool/main/a/automake-1.16/automake_1.16.1-4ubuntu3_all.deb

For opensuse, we just patch the python.m4 file.

PR #448, opened 2020-03-16
ESYS: Implement ESYS 3.0 ABI break

AndreasFuchsSIT:
Latest tpm2-tss master includes an ABI break for ESYS that affects this project in the Esys_LoadExternal() call.
See https://github.com/tpm2-software/tpm2-tss/pull/1531 for details.

Evidence that this PR works on Esys >= 3.0 can be found here: https://travis-ci.org/github/AndreasFuchsSIT/tpm2-pkcs11/builds/663011856?utm_source=github_status&utm_medium=notification
(Note that the failed jobs there originate from the pkcs11-tool problems)


williamcroberts:
Its just one build failing, the nss-tests.sh occasionally have intermittent failures in the grep command that I haven't tracked down the cause of. I re-started that job, it should pass now.

PR #462, opened 2020-03-24
Nss test workaround

williamcroberts:


williamcroberts:
Weird, the checks are not showing up for travis:
https://travis-ci.org/github/tpm2-software/tpm2-pkcs11/builds/666488343

test/integration/nss-tests.sh
```diff
@@ -23,6 +23,7 @@ function pinentry() {
   cmd="$*"
   printf 'spawn %s\nexpect "Password or Pin"\nsend -- %s\\r\nexpect eof\n' \
 	  "$cmd" "$pin" | expect
+  exit $?
```
dkobras: This change seems bogus. Now the script as a whole exists after the first call to `pinentry`. The original code looks correct to me (`return $?` is implicit in bash).

Issue #444, opened 2020-03-12
nss tests: grep fails


williamcroberts:
The nss tests keep getting intermittent failures on the grep command:
```bash
++ echo 'grep -q u,u,u on line 94 failed: 1'
grep -q u,u,u on line 94 failed: 1
```
I have no idea why this keeps happening. They were added by @dkobras, do you have any ideas?

```bash
+ echo 'Testing S/MIME certificate lookup in NSS DB via mail address'
Testing S/MIME certificate lookup in NSS DB via mail address
+ pinentry anotheruserpin certutil -L -d . --email testuser@example.org
+ pin=anotheruserpin
+ shift
+ cmd='certutil -L -d . --email testuser@example.org'
+ expect
+ printf 'spawn %s\nexpect "Password or Pin"\nsend -- %s\\r\nexpect eof\n' 'certutil -L -d . --email testuser@example.org' anotheruserpin
spawn certutil -L -d . --email testuser@example.org
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001c4) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001c4) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001e6) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001e6) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001e6) 
[...]

PR #463, opened 2020-03-25
test: fix expect script for NSS pinentry

dkobras:
Make sure all test tokens can be unlocked with matching PINs, and
propagate exit code of spawned process from expect.

Fixes: #444

### DIFF
--- a/test/integration/fixtures/client.cnf
+++ b/test/integration/fixtures/client.cnf
@@ -3,7 +3,7 @@ default_ca		= CA_default
 
 [ CA_default ]
 fixtures    = $ENV::TEST_FIXTURES
-dir			= ./
+dir			= $ENV::CA_DIR
 certs			= $dir
 crl_dir			= $dir/crl
 database		= $dir/index.txt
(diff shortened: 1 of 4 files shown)


dkobras:
Expected to fail due to #464. However, the travis builds fail because the NSS tests seem to time out, and that's certainly not expected. Alas, I cannot reproduce it locally with `.ci/travis.run`. Still investigating...

williamcroberts:
> Expected to fail due to #464. However, the travis builds fail because the NSS tests seem to time out, and that's certainly not expected. Alas, I cannot reproduce it locally with `.ci/travis.run`. Still investigating...

@dkobras can you:
1. rebase on master and force push
2. sign off your commit (git commit --amend -s)

dkobras:
> @dkobras can you:
> 
> 1. rebase on master and force push
> 2. sign off your commit (git commit --amend -s)

Both done. I've also fixed the hanging test run on the Travis infrastructure. Furthermore, the new `setup_ca` would overwrite an existing CA, leading to occasional failures when running multiple tests in parallel as done in the CI scripts. Moving each CA to a temp dir of its own fixed that as well.

williamcroberts:
@dkobras Thank you, very much appreciated!

PR #473, opened 2020-04-03
use safe arithmetic

williamcroberts:
Just to help prevent overflows in the code base, use the builtin arithmetic routines that detect overflow conditions.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -17,7 +17,7 @@ fi
 # If it's clang, enable asan
 if [[ "$CC" == clang* ]]; then
   echo "Detecting clang, enable asan"
-  export CFLAGS="-O1 -g -fsanitize=address -fno-omit-frame-pointer"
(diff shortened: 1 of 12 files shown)


williamcroberts:
dang llvm... https://bugs.llvm.org/show_bug.cgi?id=16404

williamcroberts:
I can do AC_TRY_LINK (includes, function-body, [action-if-found [, action-if-not-found]]) and if it works, do nothing, if it doesn't add to LD_FLAGS:
```
You can link with builtins by providing --rtlib=compiler-rt.

This line works for me:
./bin/clang a.cc -fsanitize=integer -rtlib=compiler-rt -lgcc_s
```

williamcroberts:
AC_LINK_IFELSE

williamcroberts:
So I wasn't able to work-around the llvm bug with configure and linking in rtlib, I would get undefined ASAN stuff. Just disable the use of the builtins when building for clang ASAN targets in the ci build.

PR #481, opened 2020-04-16
Tests: Add p11tool tests

AndreasFuchsSIT:
Ok, so this is adds a test, but only for p11tool >= 3.5.0 and < 3.6.0
Due to segfaulting on < 3.5.0 we skip.
Due to this bug of 3.6.x on clang https://gitlab.com/gnutls/gnutls/-/issues/965 we skip.

This is about the ugliest test case here, but I have no better idea...

Edit:
I now reworked this to just skip on `DOCKER_TAG=fedora-30 -a CC=clang`.

(includes #482)


AndreasFuchsSIT:
Ok, so this even hangs on a call to `p11-tool --version`. The problem is that the module definition is now global, so no idea how to handle this cases, when we cannot even test for the p11-tool's version... :-(

AndreasFuchsSIT:
Ok, so it seems to be libasan...

@williamcroberts How should I/we disable asan on this single platform for this single test ?

AndreasFuchsSIT:
I have no idea, why this piece of code does not trigger:
```sh
if [ "${DOCKER_TAG:-nodocker}" = "fedora-30" -a "${CC:-nocc}" = "clang" ]; then
  echo "p11-tool on fedora with clang is dysfunctional"
  exit 77
fi
```

@williamcroberts Can you help me here ?

williamcroberts:
> I have no idea, why this piece of code does not trigger:
> 
> ```shell
> if [ "${DOCKER_TAG:-nodocker}" = "fedora-30" -a "${CC:-nocc}" = "clang" ]; then
>   echo "p11-tool on fedora with clang is dysfunctional"
>   exit 77
> fi
> ```
> 
> @williamcroberts Can you help me here ?

Two things:
1. CC could be clang-7 or something like that, so better to check against $ASAN_ENABLED = "true"
2. DOCKER_TAG is not available in the Docker container without a tweak of the env file:
```diff
diff --git a/.ci/docker.env b/.ci/docker.env
index 9a12fb1f9325..1769cb21a472 100644
--- a/.ci/docker.env
+++ b/.ci/docker.env
@@ -12,3 +12,5 @@ TRAVIS_REPO_SLUG
 
 PYTHON_INTERPRETER
 ENABLE_COVERAGE
+
+DOCKER_TAG
```

.ci/docker-prelude.sh
```diff
@@ -15,6 +15,16 @@ get_deps "$WORKSPACE"
 export LD_LIBRARY_PATH=/usr/local/lib/
 export PATH=/root/.local/bin/:/ibmtpm974/src:$PATH
 
+# Unfortunately, p11tool unlearned the option for $HOME/.config/pkcs11/modules
+# This is true for Fedora 30 container and upstream
+# Thus we have to define it system-wide in this container
```
williamcroberts, line 20: ouch

test/integration/nss-tests.sh
```diff
@@ -76,7 +76,9 @@ echo "Initializing temporary NSS DB"
 certutil -N -d . --empty-password
 
 echo "Adding PKCS11 module in $modpath to NSS configuration"
-echo | modutil -add tpm2 -libfile "$modpath" -dbdir .
+#modutil will first ask about a running browser which we acknowledge with \n
+#sometimes it will then ask because p11-kit already knows tpm2 which we then abort with q\n
+echo -ne "\nq\n" | modutil -add tpm2 -libfile "$modpath" -dbdir .
```
williamcroberts, line 81: I hate these tools, all of them.

test/integration/p11-tool.sh.nosetup
```diff
@@ -0,0 +1,143 @@
...
+p11-kit list-modules || true
+echo "Done"
+
+echo "Should have uninitialized token"
+p11_tool --list-tokens
+if ! p11_tool --list-tokens | grep -q "Model: SW   TPM"; then
+  echo "p11tool did not find this token."
+  echo "If you are running a recent version of p11tool then you need to configure the module system-wide:"
+  echo "echo \"module: $TRAVIS_BUILD_DIR/build/src/.libs/libtpm2_pkcs11.so\" >/etc/pkcs11/modules/tpm2_pkcs11.module"
+  exit 77 #TEST SKIP
```
williamcroberts, line 68: This condition should be a skip?

test/integration/p11-tool.sh.nosetup
```diff
@@ -0,0 +1,143 @@
...
+echo "Got token url: ${TOKENURL}"
+
+echo "Testing mech and random function"
+p11_tool --list-mechanisms ${TOKENURL}
+test "$(p11tool -d 0 --generate-random=4096 ${TOKENURL} \
+            | wc -c)" == "4096"
+echo "Random function works"
+
+echo "Testing p11tool version >= 3.5.0 and (for clang) < 3.6.0 before we continue"
+#Older version of p11tool segfault during --initialize
```
williamcroberts, line 84: Maybe just skip on ASAN period, this looks brutal. If you want to keep it, more testing is more better, I wont object.

test/integration/p11-tool.sh.nosetup
AndreasFuchsSIT: Could be an ERROR (99) if you like ?

AndreasFuchsSIT: On fedora this remains a skip, on others it's now an error

test/integration/p11-tool.sh.nosetup
AndreasFuchsSIT: I have to keep it... This was a kill on ubuntu 16.04 with gcc...

AndreasFuchsSIT:
@williamcroberts So we're now skipping on fedora-30 in general. Without ASAN we don't hang anymore but the token still does not get listed. Strace shows some access, but nothing happening.
GCC works...

No idea, but so far upstream, that I don't see any alternative for now.

williamcroberts:
> @williamcroberts So we're now skipping on fedora-30 in general. Without ASAN we don't hang anymore but the token still does not get listed. Strace shows some access, but nothing happening.
> GCC works...
> 
> No idea, but so far upstream, that I don't see any alternative for now.

These tools are a bear, more coverage is better than less, but code cov shows no delta. odd....

PR #493, opened 2020-04-27
Some refactoring

AndreasFuchsSIT:
Some refactoring that should help the FAPI transition and hopefully make sense in general...

### DIFF
--- a/src/lib/token.h
+++ b/src/lib/token.h
@@ -54,8 +54,6 @@ struct sealobject {
     twist sopub;
     twist sopriv;
     twist soauthsalt;
-
-    uint32_t handle;
 };
 
 enum token_type {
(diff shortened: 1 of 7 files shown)


src/lib/token.h
```diff
@@ -54,8 +54,6 @@ struct sealobject {
     twist sopub;
     twist sopriv;
     twist soauthsalt;
-
-    uint32_t handle;
```
williamcroberts, line 58: Can you also (not in this series per-say) deal with the TODO on l75:
https://github.com/tpm2-software/tpm2-pkcs11/pull/493/files#diff-e81270e2eb0c861a9aad8a2ded26dad7R75

AndreasFuchsSIT: That's what I'm working towards with this series... :-)

AndreasFuchsSIT:
The fuzz test did not like my change.
CI should be working now.

PR #494, opened 2020-04-28
FAPI: Decoupling for esysdb & implement changeauth

AndreasFuchsSIT:
Decouple esysdb and fapi backend by enabling the union in struct token.
Moving the unsealing into the backend.
Refactor changeauth to be a backend-function.
Implement changeauth for FAPI backend.

build ontop of #493

### DIFF
--- a/src/lib/backend_fapi.c
+++ b/src/lib/backend_fapi.c
@@ -655,6 +655,7 @@ static TSS2_RC auth_cb(FAPI_CONTEXT *context, char const *description, char **au
         /* Using strstr because description may be prefixed with a crypto profile */
         if (strstr(description, at->path)) {
             /* Current FAPI falsely uses char ** instead of const char ** for return. */
+            //TODO remove strdup once we switch to FAPI 3.0
             *auth = strdup((char*) at->auth);
             if (!*auth) {
                 return TSS2_FAPI_RC_MEMORY;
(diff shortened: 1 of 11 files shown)


AndreasFuchsSIT:
Sorry, I'm leaking memory... Debugging now...

williamcroberts:
> Sorry, I'm leaking memory... Debugging now...

No worries, if you're rebased ontop of master exportr CC=clang && ./configure --enable-asan is your friend.

src/lib/backend_fapi.c
```diff
@@ -727,3 +639,225 @@ CK_RV backend_fapi_rm_tobject(token *t, tobject *tobj) {
...
+    LOGV("Searching auth value for %s", description);
+
+    struct authtable *at = (struct authtable *) userData;
+
+    for (; at->path != NULL; at = &at[1]) {
+        /* Using strstr because description may be prefixed with a crypto profile */
+        if (strstr(description, at->path)) {
+            /* Current FAPI falsely uses char ** instead of const char ** for return. */
+            //TODO remove strdup once we switch to FAPI 3.0
+            *auth = strdup((char*) at->auth);
```
williamcroberts, line 659: strdup can fail here

src/lib/backend_fapi.c
```diff
@@ -727,3 +639,225 @@ CK_RV backend_fapi_rm_tobject(token *t, tobject *tobj) {
...
+    rv = CKR_GENERAL_ERROR;
+
+    oldauth = utils_hash_pass(toldpin, user ? tok->fapi.userauthsalt : tok->fapi.soauthsalt);
+    if (!oldauth) {
+        goto out;
+    }
+
+    char label[sizeof(tok->label) + 1]; /* token-label length plus \0, cannot overflow */
+    label[sizeof(tok->label)] = '\0';
+    memcpy(&label[0], &tok->label[0], sizeof(tok->label));
```
williamcroberts, line 787: labels need to be binary safe so if your banking on a null \0, you might not get what you want if one is embedded before that.

src/lib/backend_fapi.c
```diff
@@ -727,3 +639,225 @@ CK_RV backend_fapi_rm_tobject(token *t, tobject *tobj) {
...
+
+    rc = Fapi_SetAuthCB(tok->fapi.ctx, auth_cb, &authtable[0]);
+    if (rc) {
+        LOGE("Fapi_SetAuthCB failed.");
+        goto out;
+    }
+
+    LOGV("Attempting to change auth value for %s", path);
+
+    rc = Fapi_ChangeAuth(tok->fapi.ctx, path, newauthhex);
```
williamcroberts, line 804: This seems like a lot of work just so we can FAPI_Changeauth, is their any benefit to this over just using Esys_ChangeAuth?

Im definitely more favorable to using Esys_API over mixing outside of populating data from the backends. This would just be more code to maintain for the same functionality.

src/lib/backend_fapi.c
AndreasFuchsSIT: ACK

src/lib/backend_fapi.c
AndreasFuchsSIT: Are you certain ? IMHO label might not be \0 terminated, but if there is a \0 then that is the terminator.
That's why I get the label and append a \0 to guarantee being a string...

src/lib/backend_fapi.c
AndreasFuchsSIT: The seal objects are now completely separated between ESYSDB and FAPI. This will enable us to use policies for SO or USER in the future. So I'm in favor of using this.

For the object operations themselves, I agree and I'd leave them be esys functions.

ACK ?

src/lib/backend_fapi.c
williamcroberts: > Are you certain ? IMHO label might not be \0 terminated, but if there is a \0 then that is the terminator.
> That's why I get the label and append a \0 to guarantee being a string...
You guarantee that it's null terminated, nothing more. However, the data passed is supposed to an rfc2279 string thus UTF8 compatible, but we never check. We might want to add a check in init token, that it has no embedded NULL's then we're good.

src/lib/backend_fapi.c
williamcroberts: > The seal objects are now completely separated between ESYSDB and FAPI. This will enable us to use policies for SO or USER in the future. So I'm in favor of using this.
> 
> For the object operations themselves, I agree and I'd leave them be esys functions.
> 
> ACK ?

Oh yeah I forgot about moving to policies, while I am never sure well get there, im ok with putting in some of the framing, especially since it really doesn't add complexity, it just splits out the backend for this call.

AndreasFuchsSIT:
So pending the Question of what to do with the label, I guess this is ready to be merged...
CI passes and the strdup() NULL-check was added...

williamcroberts:
> So pending the Question of what to do with the label, I guess this is ready to be merged...
> CI passes and the strdup() NULL-check was added...

That works for me for this series. Can you file a bug to scan the label for embedded NUL bytes in C_InitToken?

PR #499, opened 2020-04-29
Fix and test C_SetAttributeValue via label

AndreasFuchsSIT:
So this fixes the C_SetAttributeValue handling of the FAPI backend and adds a test using p11tool.

Problem is, that the esysdb backend fails now.
The problem is that the label is not changed.
I could trace as far as db_update_tobject_attrs() being correct; i.e. containing the correct attributes.

No idea where to search next ?

P.S. The successful runners are those that skip p11tool tests.
PP.S. The error shows in the list-all where the private key is still called "myrsakey" instead of "myrsakey2"

### DIFF
--- a/src/lib/backend.c
+++ b/src/lib/backend.c
@@ -264,7 +264,6 @@ CK_RV backend_update_tobject_attrs(token *tok, tobject *tobj, attr_list *attrs) 
 
     switch (tok->type) {
     case token_type_esysdb:
-        LOGV("Adding object to token using esysdb backend.");
         return backend_esysdb_update_tobject_attrs(tobj, attrs);
     case token_type_fapi:
         return backend_fapi_update_tobject_attrs(tok, tobj, attrs);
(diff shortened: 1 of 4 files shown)


williamcroberts:
> So this fixes the C_SetAttributeValue handling of the FAPI backend and adds a test using p11tool.
> 
> Problem is, that the esysdb backend fails now.
> The problem is that the label is not changed.
> I could trace as far as db_update_tobject_attrs() being correct; i.e. containing the correct attributes.
> 
> No idea where to search next ?
> 
> P.S. The successful runners are those that skip p11tool tests.
> PP.S. The error shows in the list-all where the private key is still called "myrsakey" instead of "myrsakey2"

I apply this asap:
```diff
diff --git a/src/lib/db.c b/src/lib/db.c
index 94841dbb17db..2145e0125603 100644
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -851,6 +851,12 @@ CK_RV db_update_tobject_attrs(unsigned id, attr_list *attrs) {
     rc = sqlite3_bind_int(stmt, 2, id);
     gotobinderror(rc, "id");
 
+    rc = sqlite3_step(stmt);
+    if (rc != SQLITE_DONE) {
+        LOGE("Could not execute stmt");
+        goto error;
+    }
+
     rc = sqlite3_finalize(stmt);
     if (rc) {
         LOGE("finalize");
```

williamcroberts:
Patch merged in https://github.com/tpm2-software/tpm2-pkcs11/pull/500

williamcroberts:
PR #501 has all these patches

PR #504, opened 2020-05-04
FAPI: SRK

AndreasFuchsSIT:
This adds the last missing piece for the fapi backend, which is SRK handling.
Thus, also changes the logic for backend selection a little, introducing a `TPM2_PKCS11_BACKEND=fapi` option.
Once https://github.com/tpm2-software/tpm2-tss/pull/1716 is done this should also pass.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,7 +35,7 @@ AC_CONFIG_FILES([Makefile lib/tpm2-pkcs11.pc])
 AC_CONFIG_HEADERS([src/lib/config.h])
 
 # test for TSS dependecies
-PKG_CHECK_MODULES([TSS2_FAPI],   [tss2-fapi])
+PKG_CHECK_MODULES([TSS2_FAPI],   [tss2-fapi >= 3.0])
 PKG_CHECK_MODULES([TSS2_ESYS],   [tss2-esys >= 2.0])
 PKG_CHECK_MODULES([TSS2_MU],     [tss2-mu])
 PKG_CHECK_MODULES([TSS2_TCTILDR], [tss2-tctildr])
--- a/src/lib/backend.c
+++ b/src/lib/backend.c
@@ -5,10 +5,25 @@
 #include "backend_esysdb.h"
 #include "backend_fapi.h"
 
-static bool is_fapi_preview_enabled(void) {
+enum backend {
+    backend_error,
+    backend_esysdb,
+    backend_fapi
+};
 
-    const char *env = getenv("FAPI_PREVIEW");
-    return env && ((!strcmp(env, "yes") || !strcmp(env, "true")));
+static enum backend get_backend(void) {
+
+    const char *env = getenv("TPM2_PKCS11_BACKEND");
+
+    if (!env || !strcasecmp(env, "esysdb")) {
+        return backend_esysdb;
+    }
+
+    if (!strcasecmp(env, "fapi")) {
+        return backend_fapi;
+    }
+
+    return backend_error;
 }
 
 /* This file includes the logic for selecting, aggregating and
@@ -25,12 +40,16 @@ static bool esysdb_init = false;
 CK_RV backend_init(void) {
     LOGV("Initializing backends");
 
-    bool is_fapi_preview = is_fapi_preview_enabled();
+    enum backend backend = get_backend();
+
+    if (backend == backend_error) {
+        return CKR_GENERAL_ERROR;
+    }
 
     CK_RV rv = backend_fapi_init();
     if (rv) {
         static const char *msg = "FAPI backend was not initialized.";
-        if (is_fapi_preview) {
+        if (backend == backend_fapi) {
             LOGE(msg);
             return rv;
         }
@@ -57,11 +76,11 @@ CK_RV backend_destroy(void) {
     LOGV("Destroying backends");
     CK_RV rv = CKR_OK;
 
-    bool is_fapi_preview = is_fapi_preview_enabled();
+    enum backend backend = get_backend();
 
     if (fapi_init) {
         rv = backend_fapi_destroy();
-        if (!is_fapi_preview) {
+        if (backend != backend_fapi) {
             rv = CKR_OK;
         }
     }
@@ -77,8 +96,9 @@ CK_RV backend_destroy(void) {
 }
 
 CK_RV backend_ctx_new(token *t) {
-    bool is_fapi_preview = is_fapi_preview_enabled();
-    if (is_fapi_preview) {
+    enum backend backend = get_backend();
+
+    if (backend == backend_fapi) {
         return backend_fapi_ctx_new(t);
     } else {
         return backend_esysdb_ctx_new(t);
@@ -104,9 +124,9 @@ void backend_ctx_free(token *t) {
  */
 CK_RV backend_create_token_seal(token *t, const twist hexwrappingkey,
                        const twist newauth, const twist newsalthex) {
+    enum backend backend = get_backend();
 
-    bool is_fapi_preview = is_fapi_preview_enabled();
-    if (is_fapi_preview) {
+    if (backend == backend_fapi) {
         if (!fapi_init) {
             LOGE("FAPI backend not initialized.");
             return CKR_GENERAL_ERROR;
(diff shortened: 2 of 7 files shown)


src/lib/backend.c
```diff
@@ -5,10 +5,45 @@
...
+    backend_unset = 0,
+    backend_error,
+    backend_esysdb,
+    backend_fapi
+};
+
+static enum backend get_backend(void) {
+
+    /* Using a static to only query env once per runtime. */
+    static enum backend backend = backend_unset;
```
williamcroberts, line 18: So ideally, someone could do:
```C
setenv("TPM2_PKCS11_BACKEND", "1", 1)
C_InitToken()
...
setenv("TPM2_PKCS11_BACKEND", "0", 1)
C_InitToken()
...
```

All in the same process and we want that to work. The query is relatively cheap, getenv is pulling a string mapped in the process memory.

src/lib/backend.c
```diff
@@ -5,10 +5,45 @@
...
+    }
+
+    env = getenv("TPM2_PKCS11_BACKEND");
+    if (env && !strcasecmp(env, "fapi")) {
+#ifdef FAPI_3_0
+        backend = backend_fapi;
+#else /* FAPI_3_0 */
+        LOGE("Using tpm2-tss (FAPI) < 3.0 does not fully work with FAPI backend. "
+             "Please use FAPI_PREVIEW=true until you can update to tpm2-tss >= 3.0.");
+        backend = backend_error;
```
williamcroberts, line 40: Just make this die in configure.

configure.ac
```diff
@@ -36,6 +36,9 @@ AC_CONFIG_HEADERS([src/lib/config.h])
 
 # test for TSS dependecies
 PKG_CHECK_MODULES([TSS2_FAPI],   [tss2-fapi])
+PKG_CHECK_MODULES([TSS2_FAPI_3_0],   [tss2-fapi >= 3.0],
+                  [AC_DEFINE([FAPI_3_0], [1], [Fapi3.0])],
+                  [AC_MSG_WARN([FAPI 3.0 features not available.])])
```
williamcroberts, line 41: This just ends up as a backend error anyways, so make configure unhappy about it

src/lib/backend_fapi.c
```diff
@@ -10,6 +10,96 @@
 FAPI_CONTEXT *fctx = NULL;
 unsigned maxobjectid = 0;
 
+#ifndef FAPI_3_0
+/* The following piece of code is here to enable a preview of the FAPI backend on systems with a
+   TSS < 3.0, and retain the current "FAPI_PREVIEW" feature. */
```
williamcroberts, line 15: Thier has never been a release with FAPI_PREVIEW and even if their was, it was always a developer preview. I see no reason to keep the old behavior.

src/lib/backend_fapi.c
```diff
@@ -68,6 +158,14 @@ static char * tss_path_from_id(unsigned id, const char *type) {
     return path;
 }
 
+static char * path_get_parent(const char *path) {
```
williamcroberts, line 161: pointer * tight to the variable... `static char *path_get_parent()` Here and throughout.

src/lib/backend_fapi.c
```diff
@@ -10,6 +10,96 @@
...
+    size_t length;
+
+    rc = Fapi_GetEsysBlob(fctx, path, &type, &data, &length);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOGE("Cannot get Esys blob for key %s", path);
+        return CKR_GENERAL_ERROR;
+    }
+
+    twist twistdata = twistbin_new(data, length);
+    Fapi_Free(data);
```
williamcroberts, line 73: Will this leak things like data from Fapi_GetEsysBlob?

src/lib/backend.c
AndreasFuchsSIT: ok

src/lib/backend.c
AndreasFuchsSIT: Well, this is the case where we are guessing the SRK to be 81000001 but in non-preview mode. So I'd like to keep this...

configure.ac
AndreasFuchsSIT: Well, this is a warning. I guess I could make this warning more elaborate.
Or should I just set it to no-problem ?

src/lib/backend_fapi.c
AndreasFuchsSIT: I think I don't understand ?

src/lib/backend_fapi.c
AndreasFuchsSIT: Are you sure ?

src/lib/backend_fapi.c
AndreasFuchsSIT: I don't understand...

src/lib/backend_fapi.c
williamcroberts: Yes, things on master can change at any time. The only thing I'll be a huge stickler on is data upgrades. We can never trash user data. But FAPI_PREVIEW was always billed as experimental and use at your own risk. Once we land a release with the feature as a first class citizen, we always need to preserve data.

src/lib/backend_fapi.c
williamcroberts: > I don't understand...

I don't understand either, I am an idoit

src/lib/backend_fapi.c
williamcroberts: Sorry mispoke, * tight to the fn name... I edited my comment to make seem more awake.

src/lib/backend.c
```diff
@@ -5,10 +5,27 @@
...
+    backend_fapi
+};
 
-    const char *env = getenv("FAPI_PREVIEW");
-    return env && ((!strcmp(env, "yes") || !strcmp(env, "true")));
+static enum backend get_backend(void) {
+
+    const char *env = getenv("TPM2_PKCS11_BACKEND");
+
+    if (!env) {
```
williamcroberts, line 18: I don't know why this block of code bothers me so much, but I am going to be pedantic.
```C
typedef enum backend backend;
enum backend {
    backend_error,
    backend_esysdb,
    backend_fapi
};

static backend get_backend(void) {

    const char *env = getenv("TPM2_PKCS11_BACKEND");

    if (!env || !strcasecmp(env, "esysdb")) {
        return backend_esysdb;
    }

    if (!strcasecmp(env, "fapi")) {
        return backend_fapi;
    }

    return backend_error;
}
```

src/lib/backend.c
```diff
@@ -25,12 +42,14 @@ static bool esysdb_init = false;
 CK_RV backend_init(void) {
     LOGV("Initializing backends");
 
-    bool is_fapi_preview = is_fapi_preview_enabled();
+    if (get_backend() == backend_error) {
```
williamcroberts, line 45: I forgot to mention this, no functions inside conditionals unless it makes the code simpler (like an fgetline loop). Here and throughout.

src/lib/backend_fapi.c
```diff
@@ -68,6 +113,14 @@ static char * tss_path_from_id(unsigned id, const char *type) {
     return path;
 }
 
+static char *path_get_parent(const char *path) {
+    char *end = rindex(path, '/');
+    if (!end) {
+        return NULL;
+    }
+    return strndup(path, end - path);
+}
```
williamcroberts, line 122: I'm going to file a bug, but I am concerned with all the path manipulation string parsing. So we need to make sure that code is robust, can we get some fuzz tests? Not necessarily now, but in the future.

src/lib/backend.c
AndreasFuchsSIT: done

src/lib/backend.c
AndreasFuchsSIT: done

src/lib/backend_fapi.c
AndreasFuchsSIT: We're actually just tokenizing stuff on '\0' whilst keeping buffer-checks.
The rest is sscanf("%08x") and the yaml-parser.
So yes, looks scary, but with simple operations.

Fuzzing is however never a bad idea...

AndreasFuchsSIT:
Since I merged https://github.com/tpm2-software/tpm2-tss/pull/1719 we should continue over at https://github.com/tpm2-software/tpm2-pkcs11/pull/505 because that uses the new callback style.

Issue #507, opened 2020-05-07
Add FAPI to changelog


williamcroberts:
could you update the CHANGELOG to include FAPI as a feature mention? Im not sure how you want to word that, so I figured best to let you do it! Thanks!

PR #508, opened 2020-05-07
CHANGELOG: Add note on backends and fapi

AndreasFuchsSIT:
Fixed: #507


williamcroberts:
Hahah you can do bitmoji links, how silly.
bitmoji

PR #509, opened 2020-05-08
Some fixes from scanbuild and coverity

AndreasFuchsSIT:
Some things are just silencing the analysis tools.

### DIFF
--- a/src/lib/attrs.c
+++ b/src/lib/attrs.c
@@ -1172,6 +1172,10 @@ CK_RV attr_list_update_entry(attr_list *attrs, CK_ATTRIBUTE_PTR untrusted_attr) 
     attr_handler2 *handler = attr_lookup(t);
 
     CK_ATTRIBUTE_PTR found = attr_get_attribute_by_type(attrs, t);
+    if (!found) {
+        LOGE("Attribute entry not found");
+        return CKR_GENERAL_ERROR;
+    }
 
     /* the existing type memory should match the expected type memory */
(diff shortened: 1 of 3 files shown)


src/lib/parser.c
```diff
@@ -249,7 +257,7 @@ bool handle_attr_event(yaml_event_t *event,
         return push_handler(state, on_seq_scalar_event);
     case YAML_SEQUENCE_END_EVENT:
         /* XXX we know that sequences never come first so the previous state (map) has the key */
-        assert(state->s);
+        assert_analyzer(state->s);
```
williamcroberts, line 260: Where is this one coming from, is --debug performed at configure time? The assert() goes away if NDEBUG is defined, and --debug control's that. Where is this getting generated?

AndreasFuchsSIT: you are correct, scan-build runs ok with --enable-debug... hmpf...

AndreasFuchsSIT:
P.S. coverity hates twist like crazy; complaining about 4 resource leaks false-positive because of the pointer-voodoo... ;-)

williamcroberts:
> P.S. coverity hates twist like crazy; complaining about 4 resource leaks false-positive because of the pointer-voodoo... ;-)

Yeah it does!

src/lib/parser.c
williamcroberts: Iv'e been burned so many times going, WTH... and then going oh yeah its not a debug build so the asserts are gone.

PR #510, opened 2020-05-11
Refactoring token ids and empty tokens plus more fapi testing

AndreasFuchsSIT:
- Unify the numberspace of tok->id between esysdb and fapi
- Move the empty token creation out of db into backend
- Add tests for multiple fapi tokens (makes used of the previous changes)

### DIFF
--- a/src/lib/backend.c
+++ b/src/lib/backend.c
@@ -153,7 +153,6 @@ CK_RV backend_create_token_seal(token *t, const twist hexwrappingkey,
  */
 CK_RV backend_get_tokens(token **tok, size_t *len) {
     CK_RV rv = CKR_GENERAL_ERROR;
-    token tmp;
 
     enum backend backend = get_backend();
 
@@ -170,11 +169,6 @@ CK_RV backend_get_tokens(token **tok, size_t *len) {
             return rv;
         }
         LOGV("Esysdb returned %zi token", *len);
-
-        /* This is used to move the empty token to the end. */
-        /* TODO: Would be better to have a nicer way of doing so. */
-        tmp = (*tok)[*len - 1];
-        *len -= 1;
     }
 
     if (fapi_init) {
@@ -187,14 +181,35 @@ CK_RV backend_get_tokens(token **tok, size_t *len) {
                 return rv;
             } else {
                 LOGW(msg);
-                rv = CKR_OK;
             }
+        }
+        LOGV("FAPI + Esysdb returned %zi token", *len);
+    }
+
+    /* -1 for starting at id 1 and -1 for the empty token */
+    if (*len >= MAX_TOKEN_CNT - 2) {
+        LOGW("Too many tokens, must have less than %d to show empty tokens", MAX_TOKEN_CNT - 1);
+        return CKR_OK;
+    }
+
+    token *t = &(*tok)[*len];
+
+    for (t->id = 1; t->id < MAX_TOKEN_CNT; t->id += 1) {
+        size_t i = 0;
+        for (; i < *len; i++) {
+            if (((*tok)[i]).id == t->id) {
+                break;
+            }
+        }
+        if (i == *len) {
+            break;
         }
     }
 
-    if (esysdb_init) {
-        (*tok)[*len] = tmp;
-        *len += 1;
+    *len += 1;
+    rv = token_min_init(t);
+    if (rv != CKR_OK) {
+        return rv;
     }
 
     LOGV("Esysdb + FAPI returned %zi token", *len);
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -333,7 +333,6 @@ CK_RV db_get_tokens(token **tok, size_t *len) {
         return rc;
     }
 
-    bool has_uninit_token = false;
     size_t row = 0;
     while ((rc = sqlite3_step(stmt)) == SQLITE_ROW) {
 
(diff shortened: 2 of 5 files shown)


src/lib/backend.c
```diff
@@ -190,11 +184,26 @@ CK_RV backend_get_tokens(token **tok, size_t *len) {
...
+        return CKR_OK;
+    }
+
+    token *t = &(*tok)[*len];
+    t->id = *len + 1;
+    for (size_t i = 0; i < *len; i++) {
+        if (((*tok)[i]).id > t->id) {
+            t->id = ((*tok)[i]).id + 1;
+        }
+    }
```
williamcroberts, line 201: So what's the purpose of this? In case FAPI and ESYSDB have token id as the same (replace the 0x80 + removed below)?

It probably is even better to update `token *slot_get_token(CK_SLOT_ID slot_id)` where we don't use token->id to map slot to token, we can just use the index location in the list. Then id is never really used for anything but perhaps debug statements and sql queries.

src/lib/db.c
```diff
@@ -408,21 +406,6 @@ CK_RV db_get_tokens(token **tok, size_t *len) {
         cnt++;
     }
 
-    /* if their was no unitialized token in the db, add it */
```
williamcroberts, line 411: How do we know an unitialized token was added to the list? We're assuming FAPI will provide one? But what if FAPI fails?

src/lib/backend.c
AndreasFuchsSIT: The problem was, that if the token id comes from esysdb and the fapi-id is just esysdb + 0x80 then we will have a collision of ids within fapi-backend if we create a second token on fapi, because esysdb does not know about fapi's used ids.

src/lib/db.c
AndreasFuchsSIT: Not fapi, but the backend.c file will provide the empty token. Could also be the token.c or slot.c but I put it into backend.c

src/lib/backend.c
williamcroberts: make it so id doesn't matter to pkcs11 internals. Then it's only consumed by backends, which is already split on the backend selector enum. Then we don't care about its value.

AndreasFuchsSIT: I pretty much do so right now by looking for an id not consumed by either of the backend.
I cannot really remove the tokenid from the esysdb backend because it's used a primary key of the table.
Or should I decouple this ?

williamcroberts: > I pretty much do so right now by looking for an id not consumed by either of the backend.
> I cannot really remove the tokenid from the esysdb backend because it's used a primary key of the table.
> Or should I decouple this ?

The latter. We should be able to keep the backend code in place as is. But this id is also used for slot to token mapping. We can decouple it there, and then collisions really don't matter. The main logic is in slot.c:
```C
token *slot_get_token(CK_SLOT_ID slot_id) {

    slot_lock();

    size_t i;
    for (i=0; i < global.token_cnt; i++) {
        token *t = &global.token[i];
        if (slot_id == t->id) {
            slot_unlock();
            return t;
        }
    }

    slot_unlock();
    return NULL;
}
```
If we just map slot id to index in token list, we should be good. It would also fix some weirdness with slots possible having gaps, if token 1, 4, 7 were to exist.

The only time collisions might matter is debug print statements, we would need to print both id and back-end selector to distinguish them in the high level pkcs11 code. The backend print statements you can increase the log level and get filename and imply backend from that.

AndreasFuchsSIT: Ok, so maybe in the mean time we could merge this PR so we have support for multiple tokens via FAPI backend (which is a current bug) and refactoring the id-code comes in the next step (after I debugged the tabrmd-problem with 0x902s) ?

williamcroberts: Theirs a bug where esysdb hands out tokid 255 and you make it 256, which will overflow a byte. IIRC correctly, I map the the session handles to tokens using the top byte as an identifier to the token. So I am leaning towards no, were swapping one bug for another.

AndreasFuchsSIT: Done

src/lib/backend.c
```diff
@@ -187,14 +181,35 @@ CK_RV backend_get_tokens(token **tok, size_t *len) {
...
             }
         }
+        LOGV("FAPI + Esysdb returned %zi token", *len);
     }
 
-    if (esysdb_init) {
-        (*tok)[*len] = tmp;
-        *len += 1;
+    /* -1 for starting at id 1 and -1 for the empty token */
+    if (*len >= MAX_TOKEN_CNT - 2) {
```
williamcroberts, line 190: This would force current users have MAX tokens from esysdb into an error state, very unlikely, I've never seen more than a handful active at a time. But the old esysdb unitialized token logic had the same bug, it tried to always add an uninitialized token if one wasn't present, and if at max count, failed.

AndreasFuchsSIT: It will deny current users with MAX tokens a new empty token; existing tokens are usable if they were usable before.

AndreasFuchsSIT:
Turning this into a draft while I find the difference between my local Fedora-30 and Ubuntu dockers and upstream...

AndreasFuchsSIT:
Ok, I traced down a bug in FAPI-backend on `tok` vs `t`.
Ready to merge now.

src/lib/backend.c
williamcroberts: oh yeah, duh, LOGW + CKR_OK

PR #511, opened 2020-05-11
Add ubuntu-20.04

tstruk:


williamcroberts:
And it looks like p11 broke as well as LD_PRELOAD for asan on executables, we need to fix up the LD_PRELOAD so we can run a command, like openssl with pkcs11 library loaded.
```
Detected clang version: 10.0.0
Looking for libasan to LD_PRELOAD at: /usr/lib/llvm-10.0/lib/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
Found libasan at: /usr/lib64/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
export LD_PRELOAD="/usr/lib64/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so"
turning off asan detection for running commands...
ERROR: ld.so: object '/usr/lib64/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
engine "pkcs11" set.
/workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so: undefined symbol: __asan_option_detect_stack_use_after_return
Unable to load module /workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so
/workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so: undefined symbol: __asan_option_detect_stack_use_after_return
Unable to load module /workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so
PKCS11_get_private_key returned NULL
cannot load Private Key from engine
140139537544512:error:81065401:libp11:pkcs11_CTX_load:Unable to load PKCS#11 module:p11_load.c:79:
140139537544512:error:26096080:engine routines:ENGINE_load_private_key:failed loading private key:../crypto/engine/eng_pkey.c:77:
unable to load Private Key
Failed to run the setup script
/P_RSA/HE:/P_RSA/HE/EK:/P_RSA/LOCKOUT:/P_RSA/HS:/P_RSA/HS/SRK
Execution of ../test/integration/nss-tests.sh failed: 1
FAIL test/integration/nss-tests.sh (exit status: 1)
```

williamcroberts:
> So scan-build found something new in db.c...

Thats been merged, ill retrigger but it looks like other things are broken too.

williamcroberts:
> > So scan-build found something new in db.c...
> 
> Thats been merged, ill retrigger but it looks like other things are broken too.

@tstruk wait you need to rebase this on top of that commit, I forgot travis doesn't do that and repush

tstruk:
rebased

williamcroberts:
@tstruk what will give us testing with openssl 1.0.2 if we drop this?

tstruk:
> @tstruk what will give us testing with openssl 1.0.2 if we drop this? I know at one point in time we need to EOL < 1.1 OSSL support, not sure if this is the best time for that. Since the LTS distro is EOL April 2021

That's why I wanted to drop 18.04 in the first place:
https://github.com/tpm2-software/tpm2-tss/pull/1723#issuecomment-626791856

and only keep the oldest and the newest.
We could also test this on Fedora using compat-openssl10
https://src.fedoraproject.org/rpms/compat-openssl10 

We would also need a new fedora-32 image, so we could update fedora-30 + compat-openssl10  and add a new fedora-32, but I don't like to have too many images, because every new image in CI adds to the already long build time.

tstruk:
> @tstruk what will give us testing with openssl 1.0.2 if we drop this? I know at one point in time we need to EOL < 1.1 OSSL support, not sure if this is the best time for that. Since the LTS distro is EOL April 2021

BTW 1.0.2 looks like we will have some work to do:
https://wiki.openssl.org/index.php/OpenSSL_3.0#Upgrading_to_OpenSSL_3.0_from_OpenSSL_1.0.2

williamcroberts:
> > @tstruk what will give us testing with openssl 1.0.2 if we drop this? I know at one point in time we need to EOL < 1.1 OSSL support, not sure if this is the best time for that. Since the LTS distro is EOL April 2021
> 
> BTW 1.0.2 looks like we will have some work to do:
> https://wiki.openssl.org/index.php/OpenSSL_3.0#Upgrading_to_OpenSSL_3.0_from_OpenSSL_1.0.2

Yeah I wanted to go look at what OSSL 3 was looking like.

williamcroberts:
> > @tstruk what will give us testing with openssl 1.0.2 if we drop this? I know at one point in time we need to EOL < 1.1 OSSL support, not sure if this is the best time for that. Since the LTS distro is EOL April 2021
> 
> That's why I wanted to drop 18.04 in the first place:
> [tpm2-software/tpm2-tss#1723 (comment)](https://github.com/tpm2-software/tpm2-tss/pull/1723#issuecomment-626791856)
> 
> and only keep the oldest and the newest.
> We could also test this on Fedora using compat-openssl10
> https://src.fedoraproject.org/rpms/compat-openssl10
> 
> We would also need a new fedora-32 image, so we could update fedora-30 + compat-openssl10 and add a new fedora-32, but I don't like to have too many images, because every new image in CI adds to the already long build time.

Yeah I agree in keeping the matrix as small as possible. Maybe for now, just add 20.04 and then we can consider how to minimize it at a later date. Perhaps just file bug so in April 2021 we kill the Ubuntu 16.04 image and officially drop OSSL 1.0.2 support.

We want to make sure we have coverage on:
- Ubuntu
- OpenSuse
- Fedora
- OSSL 1.1
- OSSL 1.0.2

I forgot that 16.04 is the last thing in the matrix on 1.0.2. I also like to have testing against the LTS versions.

How we get there I really don't care.

PR #514, opened 2020-05-13
test/pkcs11-dbup: backport GCC 10 compilation fix to tpm2-pkcs11 1.0

diabonas:
Commit 260c84e3e9a1f44b09b690933e16900459739772 fixed compilation with GCC 10, but is only available in tpm2-pkcs11 >= 1.1.0. To run the database upgrade test with GCC 10, the fix must be applied to the (otherwise EOL) version 1.0 as well. Since it is tiny, use sed instead of applying a proper patch for greater portability.

### DIFF
--- a/test/integration/pkcs11-dbup.sh.nosetup
+++ b/test/integration/pkcs11-dbup.sh.nosetup
@@ -25,12 +25,12 @@ trap onerror ERR
 #
 pushd "$tempdir"
 
-oldver=1.0
+oldver=1.0.3
 if ! wget "$PACKAGE_URL/releases/download/$oldver/tpm2-pkcs11-$oldver.tar.gz"; then
     echo "Could not download old version tpm2-pkcs11-$oldver.tar.gz"
     exit 77
 fi
-if ! sha256sum "tpm2-pkcs11-$oldver.tar.gz" | grep -q ^883d976f7f1d3c1097612615916ce4d4568545eb658f59c445ce3d475d391e9f; then
+if ! sha256sum "tpm2-pkcs11-$oldver.tar.gz" | grep -q ^6542049c0cc217b4372da52ea207fcb22e15afb2e76b9dc2f3d126b3147780c7; then
     echo "Integrity check of tpm2-pkcs11-$oldver.tar.gz failed"
     exit 99
 fi


test/integration/pkcs11-dbup.sh.nosetup
```diff
@@ -37,6 +37,11 @@ fi
 tar xzf "tpm2-pkcs11-$oldver.tar.gz"
 
 pushd "tpm2-pkcs11-$oldver"
+
+# Backport commit 260c84e3e9a1f44b09b690933e16900459739772
+# ("Drop unused variable which is causing the build to fail with current gcc10")
+sed -i 's/^} config;$/};/' src/lib/token.h
```
williamcroberts, line 43: Why don't I just do a minor release on 1.0.2 that contains this and then set oldver to 1.0.2. As long as we stay in the same minor version for oldver, the db schema is stable.

diabonas: Sure, that would obviously be the cleaner solution :+1:

williamcroberts: RC0 is rolled: https://github.com/tpm2-software/tpm2-pkcs11/releases/tag/1.0.2-RC0

williamcroberts: The release is done, but the image is updated to the newest IBM simulator which causes issues with the test suite yet again....

williamcroberts:
@diabonas https://github.com/tpm2-software/tpm2-pkcs11/releases/tag/1.0.3 is ready, do want to make that the checkout version or shall I?

diabonas:
@williamcroberts Done, test failures are due to the new simulator version (https://github.com/tpm2-software/tpm2-pkcs11/issues/525), but at least `test/integration/pkcs11-dbup.sh.nosetup` passes :)

williamcroberts:
Yep passes, will merge. Waiting to see if Ken updates that simulator for us. I need to check back on that.

Pass can be seen on [Line 2948](https://travis-ci.org/github/tpm2-software/tpm2-pkcs11/jobs/698597002#L2948)

PR #519, opened 2020-05-27
test/pkcs-get-mechanism: allow a maximum key size of 3072 bits

diabonas:
The freshly released ibmswtpm2 version 1628 increases the maximum key
size to 3072 bits, leading to a test failure:
```
[  ERROR   ] --- 0xc00 != 0x800
[   LINE   ] --- test/integration/pkcs-get-mechanism.int.c:151: error: Failure!
[  FAILED  ] test_get_mechanism_info_good
```
Accept the previous key size of 2048 bits as well for backwards compatibility with older versions of the simulator.

### DIFF
--- a/test/integration/pkcs-get-mechanism.int.c
+++ b/test/integration/pkcs-get-mechanism.int.c
@@ -148,7 +148,7 @@ void test_get_mechanism_info_good(void **state) {
         assert_int_equal(rv, CKR_OK);
 
         assert_int_equal(mech_info.ulMinKeySize, 1024);
-        assert_int_equal(mech_info.ulMaxKeySize, 2048);
+        assert_in_range(mech_info.ulMaxKeySize, 2048, 3072);
         assert_int_equal(mech_info.flags, rsa_mechs[i].flags);
     }
 


diabonas:
BTW, even with this change the new version of the simulator in its default configuration is not usable for tpm2-pkcs11 test suite: allowing RSA 3072 keys increased the object size, but the NV memory size has not been updated to reflect that. As a result, only three objects can be persisted, which is not enough for the test suite:
```
tpm2_ptool init --primary-auth=anotherpobjpin --primary-handle=0x81000002 --primary-auth=foopass --path=/tmp/tpm_server_rQCCuK
action: Added
id: 2

[...]
rsa: [...]
WARNING:esys:src/tss2-esys/api/Esys_EvictControl.c:330:Esys_EvictControl_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_EvictControl.c:114:Esys_EvictControl() Esys Finish ErrorCode (0x0000014b) 
ERROR: Esys_EvictControl(0x14B) - tpm:error(2.0): insufficient space for NV allocation
ERROR:esys:src/tss2-esys/esys_tr.c:357:Esys_TR_Close() Error: Esys handle does not exist (70018). 
ERROR: Esys_TR_Close(0x70018) - esapi:The ESYS_TR resource object is bad
ERROR: Unable to run tpm2_evictcontrol
Failed to run the setup script
```
To make the new simulator version compatible with the test suite, `NV_MEMORY_SIZE` has to be set to a suitably high value during compilation, see https://github.com/kgoldman/ibmswtpm2/pull/4.

kgold2:
diabonas is correct.  MIN_EVICT_OBJECTS should also be increased to 7 to match the PC Client specification.

PR #530, opened 2020-06-18
C_CreateObject: Support CKO_DATA object

williamcroberts:
Support storing secrets in the DB wrapped with the token wrapping key.
Since the wrapping key can unwrap all object authorizations, using the
TPM seal interface adds nothing. If you have the wrapping key, or can
brute force it, you can use the object. An object with a sealed value
means you can also release the secret. Also, a downside when using the
TPM for sealing is speed and the data cannot be bigger than 128 bytes.
Using the token wrapping key has none of these limitations.
    
Note: Only CKA_PRIVATE CK_TRUE objects are supported. Since the library
can make no guarentees about the attributes, an accidental or malicious
change to the CKA_PRIVATE value from CK_TRUE to CK_FALSE could cause
the plaintext value to be persisted to the store. Thus *ONLY* support
CKA_PRIVATE CKA_TRUE CKO_DATA objects for now.
    
Test it by using pkcs11-tool as well as a C integration test and probe
the sqlite3 db for the secret to ensure it is not leaked.

### DIFF
--- a/src/lib/attrs.c
+++ b/src/lib/attrs.c
@@ -3,6 +3,8 @@
 #include <assert.h>
 #include <stdlib.h>
 
+#include <openssl/crypto.h>
+
 #include "attrs.h"
 #include "log.h"
 #include "pkcs11.h"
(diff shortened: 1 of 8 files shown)


williamcroberts:
Hrmm, their is an issue with this, CKA_PRIVATE is mutable. The spec allows for FALSE --> TRUE but not TRUE -->FALSE, but we currently enforce that by that fact that TPM objects are immutable and the TPM is enforcing the release of the private data. In this case the secret comes back from the TPM, so if an attacker changed the value of CKA_PRIVATE from CK_TRUE to CK_FALSE && logged in that would expose the secret (as desired), and then the user called C_SetAttributeValue to change something else, that would cause the code to write the secret to the store. An undesired consequence for sure. I think for now, the smart move is to require that CKA_PRIVATE be true for CKO_DATA objects, and reject anything that has CK_FALSE.

So I don't think I can do CKA_PRIVATE := CK_FALSE objects. I'll have to tweak the code to prevent that case. No one is asking for that particular case, so their's no point in adding it.

I think if we need to perform this type of procedure, we would want to HMAC the attributes with the wrapping key, and store the HMAC, this way you can only update them if the wrapping key is released. I think we just want to add this feature for good measure anyways.

PR #545, opened 2020-07-16
interoperability: support linking in tpm2_create objects

williamcroberts:
Support:
  - tpm2_createprimary transient objects for two template friendly names
  - linking in tpm2_create key blobs.


williamcroberts:
Depends on:
https://github.com/tpm2-software/tpm2-abrmd/pull/720

Issue #538, opened 2020-06-30
set tss version to minimum required and autodetect fapi


williamcroberts:
if FAPI is present, enable the fapi backend, else don't configure it in.

PR #554, opened 2020-07-23
configure: autodetect if FAPI is enabled

williamcroberts:
Don't require that FAPI is installed by default and auto-detect it.
However, keep the requirement in the CI build so nothing unknowingly
breaks.

Fixes: #538

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,7 +35,6 @@ AC_CONFIG_FILES([Makefile lib/tpm2-pkcs11.pc])
 AC_CONFIG_HEADERS([src/lib/config.h])
 
 # test for TSS dependecies
-PKG_CHECK_MODULES([TSS2_FAPI],   [tss2-fapi >= 3.0])
 PKG_CHECK_MODULES([TSS2_ESYS],   [tss2-esys >= 2.0])
 PKG_CHECK_MODULES([TSS2_MU],     [tss2-mu])
 PKG_CHECK_MODULES([TSS2_TCTILDR], [tss2-tctildr])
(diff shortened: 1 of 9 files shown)


src/lib/token.h
```diff
@@ -91,7 +89,7 @@ struct token {
             sealobject sealobject;
         } esysdb; /* esysdb */
         struct {
-            FAPI_CONTEXT *ctx;
+            void *fapi_ctx;
```
williamcroberts, line 92: The reason I had tpm_ctx pointer is to keep the TPM types minimized from the rest of the system, hence uint32_t vs ESYS_TR

configure.ac
```diff
@@ -72,6 +71,33 @@ AC_ARG_ENABLE(
 AS_IF([test "x$esapi_sf" = "xyes"],
        [do_esapi_manage_flags])
 
+AC_ARG_WITH(
```
AndreasFuchsSIT, line 74: Shouldn't this be AC_ARG_ENABLED with auto being default ?

configure.ac
```diff
@@ -72,6 +71,33 @@ AC_ARG_ENABLE(
...
+  [fapi],
+  [AS_HELP_STRING([--with-fapi],
+    [enable or disable the fapi backend. Default is "auto" to autodetect])],
+    [with_fapi=$enableval],
+    [with_fapi=auto])
+
+AS_IF([test "x$enable_overflow" = "xno"],
```
AndreasFuchsSIT, line 81: This part I don't understand ?
Different workstream ?

williamcroberts: me either, I think it's just a copy paste error.

configure.ac
williamcroberts: I initially wen't that route, but decided to go to "with" option, since it's --with-PACKAGE or --without-PACKAGE and this completely removed the FAPI package dependency.

But I'd be open to arguments convincing me otherwise. It's also a feature... so I could see AC_ARG_ENABLED as well.

williamcroberts: I think I convinced myself to go back to AC_ARG_ENABLED

AndreasFuchsSIT:
@williamcroberts Shall we make FAPI the default backend if --enable-fapi is passed ?

williamcroberts:
> @williamcroberts Shall we make FAPI the default backend if --enable-fapi is passed ?

I already hate having this configure flags/autodetect. Every build time configure flag is another product/SKU. Which means we need to test and support the N configurations, and then train users on differing behavior and deal what one distro configures N and another M.... So i'd rather not but perhaps could be per-swayed.

Issue #546, opened 2020-07-16
tpm2_ptool: does import need alg?


williamcroberts:
Does import need --alg and why doesn't EC work?

PR #556, opened 2020-07-24
tpm2_ptool: import algorithm option updates, support ECC

williamcroberts:
- a18bf6341d94 tpm2_ptool: support ecc import
- dbb7515a19c9 tpm2_ptool: remove checks in import commandlet

Fixes: #546


williamcroberts:
Update CHANGELOG to mentions ECC import support.

PR #569, opened 2020-07-30
db: code cov increase

williamcroberts:
Add some additional db tests to cover error paths to increase code
coverage.
    
These tests specifically target:
 - _get_blob
 - get_blob
 - get_blob_null
 - db_tobject_new


williamcroberts:
Merge this and add more coverage tests.

Issue #563, opened 2020-07-30
pkcs11-tool --init core dumps or fails to work second time


bnevis-i:
## Environment

Board: APL UPSquared with PTT
OS: Ubuntu 20.04 (opensc 0.20.0-3)
tpm2-tss: 4.2.1 (tarball)
tpm2-pkcs11: 1.3.1 (tarball)
tpm2-tools: Not installed
tpm2-abrmd: Not installed

## Error message
```
pkcs11-tool --module /usr/lib/libtpm2_pkcs11.so --init-token --label tpmhsm --so-pin (redacted) --pin (redacted)
Using slot 0 with a present token (0x1)
Token successfully initialized
Segmentation fault (core dumped)

pkcs11-tool --module /usr/lib/libtpm2_pkcs11.so --init-token --label tpmhsm --so-pin (redacted) --pin (redacted)
Using slot 0 with a present token (0x1)
ERROR: step error: UNIQUE constraint failed: tokens.id
ERROR: Could not add token to db
error: PKCS11 function C_InitToken failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.
```

williamcroberts:
The return seems fine, I have set up so you cannot have N tokens with the same name in a TPM2_PKCS11_STORE. I think the abort is in pkcs11-tool.

Maybe run it valgrind or a debugger to see where it hits the abort call? I cannot reproduce.

bnevis-i:
Debugger output:
```
Using slot 0 with a present token (0x1)
Token successfully initialized

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff731e94e in twist_len (tstring=0x0) at src/lib/twist.c:157
157		size_t len = str->end - str->data;
(gdb) bt
#0  0x00007ffff731e94e in twist_len (tstring=0x0) at src/lib/twist.c:157
#1  0x00007ffff73176a5 in tpm_loadobj (ctx=0x5555555aabe0, phandle=4293479, auth=0x0, pub_data=0x0, priv_data=0x0, 
    handle=0x5555555aebc8) at src/lib/tpm.c:793
#2  0x00007ffff73106aa in session_ctx_login (ctx=0x5555555ad590, user=1, pin=0x7fffffffe39b "123456", pinlen=6)
    at src/lib/session_ctx.c:285
[...]

PR #572, opened 2020-07-31
various fixes

williamcroberts:
A command like:
pkcs11-tool --module /usr/lib/libtpm2_pkcs11.so --init-token --label tpmhsm --so-pin foo --pin bar

Will cause a C_Login even because --pin is specified. However, C_InitPIN
has not been called to initialize the userpin. This causes a NPD when
trying to load the user sealobjects public and private blobs.

Their's also a bug when a token is inserted with the same name and UNIQUE db constraint fails. This causes the cleanup code to free a mutex that is currently being held. Fix that too.

Add a check to make sure the label doesn't have an embedded null byte, per the spec.

Add tests for everything

Fixes: #563

### DIFF
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -935,7 +935,7 @@ out:
 error:
     rc = sqlite3_finalize(stmt);
     if (rc != SQLITE_OK) {
-        LOGW("Could not finalize stmt: %d", rc);
+        LOGW("Could not finalize stmt: %s", sqlite3_errmsg(global.db));
     }
 
     rollback();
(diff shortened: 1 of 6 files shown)


williamcroberts:
Looks like the behavior of the tool isn't consistent enough, will have to write C level test.

williamcroberts:
> Looks like the behavior of the tool isn't consistent enough, will have to write C level test.

Yep, test gets skipped on 16.04 on master:
  - https://travis-ci.org/github/tpm2-software/tpm2-pkcs11/jobs/713760570

PR #573, opened 2020-07-31
C_Login: test for segfault when user pin not setup

williamcroberts:
A command like:
pkcs11-tool --module /usr/lib/libtpm2_pkcs11.so --init-token --label tpmhsm --so-pin foo --pin bar

Will cause a C_Login even because --pin is specified. However, C_InitPIN
has not been called to initialize the userpin. This causes a NPD when
trying to load the user sealobjects public and private blobs.

Related-to: #563


williamcroberts:
This test gets skipped, we need to add C code test...

PR #596, opened 2020-09-03
Code cov increase

williamcroberts:


williamcroberts:
-2: this removes transactional atomicity where it matters, for when row id is to big or 0. We want to abort the transaction with the DB and rollback.

Issue #565, opened 2020-07-30
pkcs11-tools RSA keypair gen fails


bnevis-i:
## Environment

Board: APL UPSquared with PTT
OS: Ubuntu 20.04 (opensc 0.20.0-3)
tpm2-tss: 4.2.1 (tarball)
tpm2-pkcs11: 1.3.1 (tarball)
tpm2-tools: 4.1.3 (tarball)
tpm2-abrmd: Not installed
Token initialized with tpm2_ptool with label tpmhsm

## Error message
```
id=`uuidgen | tr -d -`
pkcs11-tool --module /usr/lib/libtpm2_pkcs11.so --token-label tpmhsm --login --pin (redacted) --key-type rsa:2048 --id "$id" --keypairgen --mechanism RSA-PKCS-KEY-PAIR-GEN --label rsakey
ERROR: CKA_DERIVE=true not supported
ERROR: Failed checking private attrs
error: PKCS11 function C_GenerateKeyPair failed: rv = CKR_ATTRIBUTE_VALUE_INVALID (0x13)
Aborting.

```

Equivalent functionality works on softhsm and nitrokey hsm

bnevis-i:
> I don't think the TPM can support derive. What version of pkcs11-tool are you using, CKA_DERIVE seems to be absent from the template on all the versions we have tested on. It seems to be opt-in via the `--derive` option.

pkcs11-tool is the version from opensc 0.20.0-3 in Ubuntu 20.04.  I have also tried setting OPENSC_DEBUG=9 and passing in --verbose flags, to no avail.  The first two ERROR lines are in libtpm2_pkcs11.so and the "error:" and "Aborting" match strings in /usr/bin/pkcs11-tool.

williamcroberts:
We run our CI on ubuntu 20.04 and the pkcs11-tool seems to be creating RSA keys just fine... im not sure why it's adding derive flag... For example: https://travis-ci.org/github/tpm2-software/tpm2-pkcs11/jobs/712261942#L3176

We use the below for generating rsa keys:
`pkcs11-tool -m <module> --slot-index=0 --label="myrsakey" --login --pin=mynewuserpin --keypairgen`

I wonder if it's becuase of the different key specifiers or mechanism.... they're fickle tools at times.

bnevis-i:
With update TSS library and command line above, with a firmware TPM, I get:

```
Using slot 0 with a present token (0x1)
[...]

PR #598, opened 2020-09-03
Ignore wrap unwrap key attrs

williamcroberts:
Fix keypair generation for pkcs11-tool when key-type is used:
```bash
tpm2pkcs11-tool --token-label=label --login --pin=myuserpin --keypairgen --label rsakey4 --key-type=rsa:1024
```

 Fixes: #565


bnevis-i:
LGTM.  Verified this PR allows pkcs11-tool to create a keypair.

PR #606, opened 2020-09-30
tpm2_ptool: prefer 0x81000001 to store primary key objects

diabonas:
According to the [TCG TPM v2.0 Provisioning Guidance](https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf#page=29), the SRK should be stored at 0x81000001. Check whether this handle is available and store the primary key object there during `tpm2_ptool init`, falling back to the current behaviour of looking for the first available handle if it is already occupied.

### DIFF
--- a/tools/tpm2_pkcs11/commandlets_store.py
+++ b/tools/tpm2_pkcs11/commandlets_store.py
@@ -99,7 +99,14 @@ class InitCommand(Command):
                     if not use_existing_primary:
                         pobj_ctx = create_primary(tpm2, hierarchyauth, pobjauth, transient_parent)
                         if not transient_parent:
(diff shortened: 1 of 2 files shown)


williamcroberts (approved):
Yeah looks like I broke that huh? It used to default the primary-handle option to 0x81000001 IIRC

williamcroberts:
> Yeah looks like I broke that huh? It used to default the primary-handle option to 0x81000001 IIRC

oh nevermind, I see the difference upon refreshing my memory. --primary-handle defaults to 0x81000001, as in, it's already there, just use this address, else it will create and persist a primary object at the first available index, which may not be 0x81000001.

PR #609, opened 2020-10-15
ssl_util: let openssl verify output buffer length

williamcroberts:
When calling EVP_PKEY_encrypt, let OpenSSL determine if the buffer
length is too small. Per the manpage, the required minimum output buffer
sizes are quite nuanced, and I quote from:
  - https://www.openssl.org/docs/man1.1.1/man3/RSA_public_encrypt.html

"to must point to a memory section large enough to hold the maximal possible decrypted data (which is equal to RSA_size(rsa) for RSA_NO_PADDING, RSA_size(rsa) - 11 for the PKCS #1 v1.5 based padding modes and RSA_size(rsa) - 42 for RSA_PKCS1_OAEP_PADDING). padding is the padding mode that was used to encrypt the data. to and from may overlap."

### DIFF
--- a/src/lib/ssl_util.c
+++ b/src/lib/ssl_util.c
@@ -117,7 +117,8 @@ EC_KEY *EVP_PKEY_get0_EC_KEY(EVP_PKEY *pkey) {
 
     return pkey->pkey.ec;
 }
-
+#else
+#include <openssl/evperr.h>
 #endif
 
 static CK_RV convert_pubkey_RSA(RSA **outkey, attr_list *attrs) {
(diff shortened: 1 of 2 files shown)


src/lib/ssl_util.c
```diff
@@ -355,8 +344,15 @@ CK_RV ssl_util_encrypt(EVP_PKEY *pkey,
 
     size_t outlen = *ctextlen;
     rc = EVP_PKEY_encrypt(pkey_ctx, ctext, &outlen, ptext, ptextlen);
-    if (!rc) {
-        SSL_UTIL_LOGE("Could not perform RSA public encrypt");
```
williamcroberts, line 359: This patch exposes an issue.. where:
ERROR on line: "355" in file: "src/lib/ssl_util.c": Could not perform RSA public encrypt: error:0406D06E:rsa routines:RSA_padding_add_PKCS1_type_2:data too large for key size

EVP_PKEY_encrypt can return negative values, and they we're being treated as success.  I will have to dig into this more.

Issue #567, opened 2020-07-30
Out of memory for object context when doing simple RSA key operations in Python


bnevis-i:
## Environment

Board: APL UPSquared with PTT
OS: Ubuntu 20.04 (opensc 0.20.0-3)
tpm2-tss: 4.2.1 (tarball)
tpm2-pkcs11: 1.3.1 (tarball)
tpm2-tools: 4.1.3 (tarball)
tpm2-abrmd: Not installed
Token initialized with tpm2_ptool with label tpmhsm

RSA keypair generated with
`tpm2_ptool addkey --id  (someid) --userpin (redacted) --algorithm rsa2048 --label tpmhsm --key-label rsakey

## Script
```
import os
import pkcs11

lib = pkcs11.lib('libtpm2_pkcs11.so')

token = lib.get_token(token_label='tpmhsm')

with token.open(rw=False, user_pin='123456') as session:
    print(session)

    public = session.get_key(label='rsakey',object_class=pkcs11.constants.ObjectClass.PUBLIC_KEY)
    print(public)

    private = session.get_key(label='rsakey',object_class=pkcs11.constants.ObjectClass.PRIVATE_KEY)
    print(private)

    plaintext = '1234567890123456'

    ciphertext = public.encrypt(plaintext, mechanism=pkcs11.mechanisms.Mechanism.RSA_PKCS)
    print(ciphertext)

    plaintext = 'notdecrypted'
    plaintext = private.decrypt(ciphertext, mechanism=pkcs11.mechanisms.Mechanism.RSA_PKCS)
    print(plaintext)
```

## Error message
```
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001c4) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001c4) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
[...]

PR #610, opened 2020-10-16
python-pkcs11: fix 5.2 style returns and add tests

williamcroberts:
- Fix issue #567, C_Decrypt wasn't doing the 5.2 style returns for public keys when getting output buffer size, so fix it.
- Add a C integration test for that bug
- Add a python-pkcs11 integration test for that bug

Fixes: #567

### DIFF
--- a/Makefile-integration.am
+++ b/Makefile-integration.am
@@ -9,7 +9,8 @@ integration_scripts = \
     test/integration/openssl.sh \
     test/integration/pkcs11-javarunner.sh.java \
     test/integration/nss-tests.sh \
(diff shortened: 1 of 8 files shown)


williamcroberts:
@bnevis-i here is the PR.

Note this will fail until the containers get updated:
  - https://github.com/tpm2-software/tpm2-software-container/pull/40

bnevis-i (approved):

williamcroberts:
https://github.com/tpm2-software/tpm2-software-container/pull/40 merged, waiting on Dockerhub rebuild to re-spin CI.

PR #615, opened 2020-10-27
test: add support for swtpm simulator

diabonas:
tpm2-tss supports the [swtpm](https://github.com/stefanberger/swtpm) TPM simulator since version 3.0.0. One small change in [` test/integration/pkcs-get-mechanism.int.c`](https://github.com/tpm2-software/tpm2-pkcs11/compare/master...diabonas:add-swtpm?expand=1#diff-c9fd97e9e9ba72d316720e723fcea5c104ebb2e8b3810e49c401d8045725769a) was necessary because swtpm supports more elliptic curves than ibmswtpm, other than that everything worked out of the box.

Support for ibmswtpm (aka `tpm_server`) is kept for backwards compatibility, but swtpm is preferred if it is available. Since the Docker images come with swtpm preinstalled, this means that this PR switches CI testing from ibmswtpm to swtpm as well.

### DIFF
--- a/.ci/docker-prelude.sh
+++ b/.ci/docker-prelude.sh
@@ -13,7 +13,6 @@ source $TRAVIS_BUILD_DIR/.ci/download-deps.sh
 get_deps "$WORKSPACE"
 
 export LD_LIBRARY_PATH=/usr/local/lib/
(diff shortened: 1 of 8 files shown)


williamcroberts (approved):
Thanks @diabonas

PR #634, opened 2020-12-16
[WIP] Add sshd hostkey documentation

whooo:
As discussed during a tpm.dev meeting here is WIP documentation for using tpm2-pkcs11 and ssh-agent for sshd host keys.
I did very basic testing now, but I have previously used a very similar setup (but with another PKCS11 module).

### DIFF
--- a/docs/SSH_HOSTKEYS.md
+++ b/docs/SSH_HOSTKEYS.md
@@ -1,0 +1,46 @@
+# SSH host keys configuration
+
+This example will go through using a TPM 2.0 for the openssh sshd host keys.
+As the ssh daemon supports using the SSH agent protocol we will use tpm2-pkcs11 with an ssh-agent.
+
+# Create a store and key
+The PIN is currently needed as ssh-agent doesn't seem to support an empty string PIN (just no PIN at all)
+```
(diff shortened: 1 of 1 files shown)


docs/SSH_HOSTKEYS.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Add the token to the agent
+```
+# SSH_AUTH_SOCK=/tmp/hostagent.sock ssh-add -s /usr/lib/x86_64-linux-gnu/libtpm2_pkcs11.so
+```
+
+# Get the public key in SSH format
+```
+# SSH_AUTH_SOCK=/tmp/hostagent.sock ssh-add -L
+ssh-rsa ....
```
williamcroberts, line 30: What's with the `....` is their arguments to the command that should go there?

whooo: It's the output from ssh-add -L, I'll add a note about it

whooo: @williamcroberts, I have updated it, please check it looks good

williamcroberts (approved):

Issue #632, opened 2020-12-04
Depends on tss2-esys >= 2.4


dkobras:
Since the call to `Esys_TR_GetTpmHandle()` was added to `tpm_deserialize_handle()` in commit `5b379c0d3996152dfd6f9633b59c2e04e461e038`, compiling `src/lib/tpm.c` fails with versions of `tss2-esys` before 2.4, whereas the documentation and the configure checks still specify a minimum version of 2.0.

Looking at the code, the function is only used if `tpm_deserialize_handle()` is called with a non-zero `tpmHandle` argument. At the moment, this is only true for the call in `backend_fapi.c`. In other words, rather than just bumping the minimum required version, we might also refactor the code to only include the function call when building with FAPI support, and leave minimum requirements for non-FAPI builds at their currently documented versions.

I'm happy to provide a patch either way, but would appreciate to hear your preference on handling this situation.

(We care about this issue because we'd like to roll out `tpm2-pkcs11` on machines running Ubuntu's current LTS release 20.04, which already seems to package sufficiently recent versions for all dependencies, apart from `tss2-esys` 2.4. Not having to upgrade distro-provided libs would save us a lot of effort.)

williamcroberts:
Yeah I prefer things to work with the smallest version possible, so I would take that patch to require a lower version when disabling FAPI backend.

dkobras:
I finally got around to look into this. I'm not 100% happy with #662, and would have preferred to restrict the 2.4-only call to `backend_fapi.c` itself, but this not only went against current abstractions in the code, but also created a mess with the mock-up functions for the unit tests. Hence, I went with plan B that does the job without adding too much complexity. With this change, building a working tpm2-pkcs11 module on a current Ubuntu LTS with only packaged dependencies is back to a simple `./configure && make`.

williamcroberts:
@dkobras I'm assuming you'd like a release off of master for a fix release to just package?

dkobras:
> @dkobras I'm assuming you'd like a release off of master for a fix release to just package?

Thanks! Internally, we've already had a workaround in place. So we're in no hurry, and can just drop our patch once the fix is included in a regular release.

PR #662, opened 2021-02-24
Relax tss2-esys versioned deps for non-FAPI builds

dkobras:
Commit 5b379c0d introduced a call to Esys_TR_GetTpmHandle() into a common code
path.  This function was only added to tss2-esys in version 2.4, and isn't
readily available in packaged distributions, yet. Only the FAPI backend
actually makes use of the function. Hence, splice out the call to a separate
function that can be conditionally included. (No matching stub is added for
tss2-esys < 2.4, so dependency errors can be caught already at compile time.)
This change restricts the versioned dependency on tss2-esys >= 2.4 to
FAPI-enabled builds, and thus facilitates standard builds of tpm2-pkcs11 on
current distributions.

Fixes: #632

Signed-off-by: Daniel Kobras <kobras@puzzle-itc.de>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,6 +48,10 @@ AC_DEFUN([do_esapi_manage_flags], [
 # Check for ESYS version below 2.2.1 which requires us to manage ESYS session flags
 PKG_CHECK_EXISTS([tss2-esys < 2.2.1], [do_esapi_manage_flags])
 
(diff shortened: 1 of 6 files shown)


williamcroberts (approved):
This looks fine to me and very reasonable approach I think.

PR #667, opened 2021-03-17
Support Large Buffers AES

williamcroberts:
Support Large AES buffers as well as buffers not on block boundaries.


williamcroberts:
LGTM will pass post this merge.

PR #669, opened 2021-03-23
backed/fapi: drop Esys_TR_GetTpmHandle

williamcroberts:
Since the internals of the code were changed to use ESYS_TRs fully for primary
objects, to get name checking, thier was no longer a need for calls to
convert ESYS_TR's to TPM2_HANDLE's. Thus we can drop
Esys_TR_GetTpmHandle. However, the pid value is now currently unused by
the FAPI backend, so perhaps some population values should go there?
Perhaps just the ESYS_TR?


AndreasFuchsTPM (approved):
LGTM.

Regarding the pid, I have no idea. I guess that would work.

williamcroberts:
@dkobras this gets rid of the esapi requirement even further.

PR #683, opened 2021-05-20
utils: remove debug log message from str_padded_copy

niooss-ledger:
Commit 9af101b54035 ("utils: fix stringop-overread in str_padded_copy") introduced a log message which is displayed every time the function `str_padded_copy` is used. This looks like a debug message used while working on the patch. Now that it is applied, remove the log message.


williamcroberts:
Whoops, thought I got those all.

PR #689, opened 2021-06-03
Enhanced ECDSA, DB upgrade bug fix and error message fix

williamcroberts:


src/lib/db.c
```diff
@@ -1823,6 +1823,236 @@ static CK_RV dbup_handler_from_4_to_5(sqlite3 *updb) {
...
+        CK_MECHANISM_TYPE old_mech = ((CK_MECHANISM_TYPE_PTR)(a->pValue))[i];
+        if (old_mech == CKM_ECDSA_SHA256 ||
+                old_mech == CKM_ECDSA_SHA384 ||
+                old_mech == CKM_ECDSA_SHA512) {
+            /*
+             * don't add the ones were adding, just to make the add below
+             * unconditional. This is an unlikely case as someone would have
+             * had to twiddle their object config by hand.
+             */
+            continue;
```
niooss-ledger, line 1868: This `continue;` while `i` is still incremented seems strange. This means if the condition is true, `new_mechs` will contain an entry with value `0`. Wouldn't it be better if the entry was simply skipped? (For example with new variables `new_i` and `new_mech_num` which would track the state of `new_mechs`)

williamcroberts: Yep!

PR #684, opened 2021-05-24
Test sshkeygen import

williamcroberts:
@niooss-ledger

PR #691, opened 2021-06-07
docs/BUILDING: document how to run tests in a container

niooss-ledger:
Hello,

Finding out all the dependencies which are required to run every integration test is quite painful. This PR documents how to use containers to launch tests, which helps maintaining the dependencies list.

Add a part to the `BUILDING` documentation which describes how to launch an Arch Linux and a Debian containers to install dependencies and run all tests. This also greatly helps setting up a development system for tpm2-pkcs11.

(This PR is a cleaned-up version of what I posted a few days ago when I tested another PR, https://github.com/tpm2-software/tpm2-pkcs11/pull/684#pullrequestreview-667972361)

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -88,3 +88,32 @@ To enable testing, run `make check`.
 
 **Note:** If make check runs 0 tests, you likely need the configure options `--enable-unit` and `--enable-integration`. See [Configure Options](#configure-options)
 for more details.
+
+## Running tests in a container
+
+Sometimes it is useful to be able to run tests in a fresh environment where everything is configured by default.
+Using a container is a light way to ensure the environment of the test is quite clean.
+There are several containers provided in [tpm2-software-container](https://github.com/tpm2-software/tpm2-software-container) project.
+
+For example here is a command which builds tpm2-pkcs11 and runs continuous integration tests in a Fedora 32 container with [podman](https://podman.io/):
+
+```sh
+podman run --rm -v "$(pwd):/workspace/tpm2-pkcs11" --env-file .ci/docker.env \
+    --workdir /workspace/tpm2-pkcs11 -ti ghcr.io/tpm2-software/fedora-32 \
+    bash -c .ci/docker.run
+```
+
+Building the project and running tests can be also done with specific shell commands inside a container:
+
+```sh
+# Configure tpm2-pkcs11 to build unit and integration tests
+./bootstrap
+./configure --enable-esapi-session-manage-flags --enable-fapi --enable-unit --enable-integration
+
+# Build the project and run tests
+make -j $(nproc)
+make check
+
+# Run a single test, for example to debug a failure
+make check TESTS=test/integration/pkcs-get-mechanism.int
+```
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -95,9 +95,16 @@ Sometimes it is useful to be able to run tests in a fresh environment where ever
 Using a container is a light way to ensure the environment of the test is quite clean.
 There are several containers provided in [tpm2-software-container](https://github.com/tpm2-software/tpm2-software-container) project.
 
-For example here is a command which builds tpm2-pkcs11 and runs continuous integration tests in a Fedora 32 container with [podman](https://podman.io/):
+For example here is a command which builds tpm2-pkcs11 and runs continuous integration tests in a Fedora 32 container with [Docker](https://www.docker.com/) or [Podman](https://podman.io/):
 
 ```sh
+sudo docker run --rm -v "$(pwd):/workspace/tpm2-pkcs11" --env-file .ci/docker.env \
+    --workdir /workspace/tpm2-pkcs11 -ti ghcr.io/tpm2-software/fedora-32 \
+    bash -c .ci/docker.run
+# NB. sudo can be omitted if the user is added to the "docker" group, but this
+# configuration impacts security in your system, as documented in
+# https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user
+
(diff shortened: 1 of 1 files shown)


williamcroberts (changes requested):
I don't want to maintain this, can't we just point folks to the tpm2-software-containers and how to run with docker?

williamcroberts:
I have this in my bashrc:
```bash
docker_pull() {
        docker pull ghcr.io/tpm2-software/$1
}

_docker_run() {
  local b=$(basename $(pwd)) 
  #docker run --cap-add=SYS_PTRACE --ulimit core=0 -it --env-file .ci/docker.env -v "`pwd`:/workspace/$b" "$1" /bin/bash
  docker run --cap-add=SYS_PTRACE --memory=14GB --cpus=2 -it --env-file .ci/docker.env -v "`pwd`:/workspace/$b" "$1" /bin/bash
}

docker_run() {
  local r="ghcr.io/tpm2-software/${1:-ubuntu-16.04}"
  _docker_run "$r"
}
```
Then I just `docker_pull fedora-32` and then after I have a git clone of whatever tpm2-software project I want (most support this), then I just `docker_run fedroa-32` while my current directory is that project. For instance, I have tpm2-pkcs11 cloned, I just `cd tpm2-pkcs11` and `docker_run <image>` From the bash prompt, `/workspace/tpm2-pkcs11/.ci/docker/.run`.

niooss-ledger:
@williamcroberts I have updated the PR to reference tpm2-software-container. Your bashrc configuration is useful to understand how to launch the CI tests locally, and I "translated" them into documentation so that new users can be aware that there exists a simple way to run the CI tests. I also documented some commands which enables running some tests directly, which are very useful when some failure happens.

docs/BUILDING.md
```diff
@@ -88,3 +88,32 @@ To enable testing, run `make check`.
...
+    --workdir /workspace/tpm2-pkcs11 -ti ghcr.io/tpm2-software/fedora-32 \
+    bash -c .ci/docker.run
+```
+
+Building the project and running tests can be also done with specific shell commands inside a container:
+
+```sh
+# Configure tpm2-pkcs11 to build unit and integration tests
+./bootstrap
+./configure --enable-esapi-session-manage-flags --enable-fapi --enable-unit --enable-integration
```
williamcroberts, line 111: Why `--enable-esapi-session-manage-flags`, it's usually better to let it detect the ESAPI version and autoconfigure itself.

docs/BUILDING.md
```diff
@@ -88,3 +88,32 @@ To enable testing, run `make check`.
...
+## Running tests in a container
+
+Sometimes it is useful to be able to run tests in a fresh environment where everything is configured by default.
+Using a container is a light way to ensure the environment of the test is quite clean.
+There are several containers provided in [tpm2-software-container](https://github.com/tpm2-software/tpm2-software-container) project.
+
+For example here is a command which builds tpm2-pkcs11 and runs continuous integration tests in a Fedora 32 container with [podman](https://podman.io/):
+
+```sh
+podman run --rm -v "$(pwd):/workspace/tpm2-pkcs11" --env-file .ci/docker.env \
```
williamcroberts, line 101: Why podman and not docker, or provide both? Are they just "drop in" replacements like clang and gcc? I've never bumped into podman, so I'd like to document to the tooling with the larger user base.

docs/BUILDING.md
niooss-ledger: > Why `--enable-esapi-session-manage-flags`

The only reason why I used this flag is because it is used in `.ci/docker.run`, in https://github.com/tpm2-software/tpm2-pkcs11/blob/c2d53cc1af6b9df13c832715442853b21048c273/.ci/docker.run#L67 and https://github.com/tpm2-software/tpm2-pkcs11/blob/c2d53cc1af6b9df13c832715442853b21048c273/.ci/docker.run#L86

docs/BUILDING.md
niooss-ledger: I am probably biased by the tweet from https://twitter.com/rhatdan (a Red Hat employee who contributed a lot to SELinux and Docker before working on OCI Containers and Podman, Buildah, etc.).

Personally I do not like the concept of "being a member of the `docker` group is exactly the same as being `root`", which is a major drawback of Docker. Instead of using virtual machines to contain Docker, I prefer using a solution which leverages user namespaces so that running containers does not enable running things as root. Podman exists since 2018 so it is quite recent, but with Red Hat pushing for it (with some systemd integration for example) and with the fact that its command-line interface is compatible with Docker's one, I feel like it can be documented as an alternative to Docker.

Anyway, I can add a command-line with `sudo docker` too, to emphasize that being able to launch a container with `docker` is really like being able to run any command as `root`. Would you agree with this?

niooss-ledger: > You don't need sudo docker, just docker

No. If `docker` can be used, it can be a security issue, as documented in https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user :

    The docker group grants privileges equivalent to the root user.
    For details on how this impacts security in your system, see Docker Daemon Attack Surface.

I do not want to advocate a weak security configuration without a clear warning that this is a weak security configuration and advice about stronger alternatives.

niooss-ledger: I added a message about this:

```sh
sudo docker run --rm -v "$(pwd):/workspace/tpm2-pkcs11" --env-file .ci/docker.env \
    --workdir /workspace/tpm2-pkcs11 -ti ghcr.io/tpm2-software/fedora-32 \
    bash -c .ci/docker.run
# NB. sudo can be omitted if the user is added to the "docker" group, but this
# configuration impacts security in your system, as documented in
# https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user
```

Would such a message suits you?

williamcroberts: Thats fine

PR #693, opened 2021-06-08
general: fix parse_lib_version

niooss-ledger:
`parse_lib_version` only supports MAJOR.MINOR versions, not MAJOR.MINOR.PATCH ones. This leads `pkcs11-tool -I` to display a library version which is 0.0, when using tpm2-pkcs11 1.6.0:

    $ TPM2_PKCS11_LOG_LEVEL=1 pkcs11-tool --module /usr/lib64/pkcs11/libtpm2_pkcs11.so -I
    WARNING: Could not strtoul(6.0): Success
    Cryptoki version 2.40
    Manufacturer     tpm2-software.github.io
    Library          TPM2.0 Cryptoki (ver 0.0)
    Using slot 0 with a present token (0x1)

Note the warning about `strtoul(6.0)`: this is comes from `strtoul(minor_str, &endptr, 10)` where `minor_str` is the part after the major version number in `1.6.0`.

Fix this by allowing `endptr` to be `'.'` after this `strtoul`.

Add a regression tests which detects that a zero version was returned by `pkcs11-tool`. Before this commit, the new test failed with:

    Check library line from --show-info
    WARNING: Could not strtoul(6.0-37-g854f0cc): Success
    Using slot 0 with a present token (0x1)
    pkcs11-tool reported an erroneous library version:
    Cryptoki version 2.40
    Manufacturer     tpm2-software.github.io
    Library          TPM2.0 Cryptoki (ver 0.0)

### DIFF
--- a/src/lib/general.c
+++ b/src/lib/general.c
@@ -59,7 +59,7 @@ static void parse_lib_version(CK_BYTE *major, CK_BYTE *minor) {
 
     endptr = NULL;
     val = strtoul(minor_str, &endptr, 10);
-    if (errno != 0 || endptr[0] || val > UINT8_MAX) {
+    if (errno != 0 || (endptr[0] != '\0' && endptr[0] != '.') || val > UINT8_MAX) {
         LOGW("Could not strtoul(%s): %s", minor_str, strerror(errno));
         *major = *minor = 0;
         return;
--- a/test/integration/pkcs-misc.int.c
+++ b/test/integration/pkcs-misc.int.c
@@ -144,7 +144,7 @@ static void parse_lib_version(CK_BYTE *major, CK_BYTE *minor) {
 
     endptr = NULL;
     val = strtoul(minor_str, &endptr, 10);
-    if (errno != 0 || endptr[0] || val > UINT8_MAX) {
+    if (errno != 0 || (endptr[0] != '\0' && endptr[0] != '.') || val > UINT8_MAX) {
         *major = *minor = 0;
         return;
     }
@@ -174,6 +174,7 @@ static void test_get_info(void **state) {
     CK_BYTE major;
     CK_BYTE minor;
     parse_lib_version(&major, &minor);
+    assert_int_not_equal(major | minor, 0);
     assert_int_equal(info.libraryVersion.major, major);
     assert_int_equal(info.libraryVersion.minor, minor);
 }
--- a/test/integration/pkcs11-tool.sh
+++ b/test/integration/pkcs11-tool.sh
@@ -22,6 +22,19 @@ pkcs11_tool() {
   return $?
 }
 
+echo "Check library line from --show-info"
+pkcs11_info="$(pkcs11_tool --show-info)"
+if ! printf "%s" "$pkcs11_info" | grep -q "^Library\s\+TPM2.0 Cryptoki " ; then
+    echo "pkcs11-tool did not report using TPM2.0 Cryptoki library:"
+    printf "%s\n" "$pkcs11_info"
+    exit 1
+fi
+if ! printf "%s" "$pkcs11_info" | grep -q "^Library\s\+TPM2.0 Cryptoki (ver [^0][0-9]*\." ; then
+    echo "pkcs11-tool reported an erroneous library version:"
+    printf "%s\n" "$pkcs11_info"
+    exit 1
+fi
+
 echo "Finding cert"
 id=$(pkcs11_tool --label label --list-objects --type cert | grep 'ID:' | cut -d':' -f2- | sed s/' '//g)
 echo "Found cert: $id"


src/lib/general.c
```diff
@@ -59,7 +59,7 @@ static void parse_lib_version(CK_BYTE *major, CK_BYTE *minor) {
 
     endptr = NULL;
     val = strtoul(minor_str, &endptr, 10);
-    if (errno != 0 || endptr[0] || val > UINT8_MAX) {
```
williamcroberts, line 62: I think the logic needs to be fixed around line 38, null terminate the string on the next `.` or `-` character if found.

williamcroberts: Not tested at all, but offhand I am thinking something like this:
```diff
diff --git a/src/lib/general.c b/src/lib/general.c
index eaddaf82a97a..91c3d841a21d 100644
--- a/src/lib/general.c
+++ b/src/lib/general.c
@@ -38,6 +38,13 @@ static void parse_lib_version(CK_BYTE *major, CK_BYTE *minor) {
     if (split) {
         split[0] = '\0';
         minor_str = &split[1];
+        split = strchr(minor_str, '.');
+        if (!split) {
+            split = strchr(minor_str, '-');
+        }
+        if (split) {
+            split[0]='\0';
+        }
     }
 
     if (!major_str[0] || !minor_str[0]) {
```

niooss-ledger:
> See
> ```text
>  Check library line from --show-info
>  WARNING: Could not strtoul(1f833dd): Success
>  Using slot 0 with a present token (0x1)
>  pkcs11-tool reported an erroneous library version:
>  ```

Actually the issue is not that the version contains a `-` character, but that the version is only a commit hash, which is `1f833dd`. You can see this for example in https://github.com/tpm2-software/tpm2-pkcs11/runs/2778881483?check_suite_focus=true#step:3:2919 :
```sh
tardir=tpm2-pkcs11-1f833dd && ${TAR-tar} chof - "$tardir" | eval GZIP= gzip --best -c >tpm2-pkcs11-1f833dd.tar.gz
```

In GitHub Actions, the repository is actually checkout out with:

```sh
# https://github.com/tpm2-software/tpm2-pkcs11/runs/2778881483?check_suite_focus=true#step:2:18
/usr/bin/git init /home/runner/work/tpm2-pkcs11/tpm2-pkcs11

# https://github.com/tpm2-software/tpm2-pkcs11/runs/2778881483?check_suite_focus=true#step:2:40
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules \
  --depth=1 origin +1f833dd97b4abe2fb2b4de339750056d9d2cef05:refs/remotes/pull/693/merge
```

and in the end (https://github.com/tpm2-software/tpm2-pkcs11/runs/2778881483?check_suite_focus=true#step:2:377):

```text
  HEAD is now at 1f833dd Merge ff7147658d3ada20aaa899ed7dd239c100b65c27 into 06ca8f4f0d7d11f877cfe08a0a6d1f95f52ebee4
```

As no tag was downloaded, `git describe` only returns the commit hash.

Now, the question is: how should tpm2-pkcs11's code handle the case where the git repository was cloned without any git tag?

williamcroberts:
Ahh shallow clone issue, I was wondering exactly what the root problem was you were fixing. My first instinct is to just not shallow clone (change in the checkout action in github). I think we just assume "dev" version on shallow clone. To build a valid release tarball, you'll need a full clone. My initial reaction was to do with --> depth: 0 in the github actions, but i like the speed pick up of a shallow clone (not that this project has a super large history).

I think a fix like this is much simpler:
```bash
diff --git a/src/lib/general.c b/src/lib/general.c
index eaddaf82a97a..ba1702856077 100644
--- a/src/lib/general.c
+++ b/src/lib/general.c
@@ -38,6 +38,10 @@ static void parse_lib_version(CK_BYTE *major, CK_BYTE *minor) {
     if (split) {
         split[0] = '\0';
         minor_str = &split[1];
+    } else {
+        *major = *minor = 0;
+        LOGW("Not a valid version string X.X.*");
+        return;
     }
```

Offhand I think the code right below it is dead/wrong:
```
    if (!major_str[0] || !minor_str[0]) {
        *major = *minor = 0;
        return;
    }
```

major_str[0] cannot be a null byte, minor string I guess could be a null byte, same with minor string. I think with this patch, that hunk can be safely dropped.

williamcroberts:
> `parse_lib_version` only supports MAJOR.MINOR versions, not MAJOR.MINOR.PATCH ones. This leads `pkcs11-tool -I` to display a library version which is 0.0, when using tpm2-pkcs11 1.6.0:
> 

So is this is the issue (how did everyother release work?) or is it the shallow clone and getting just just a truncated sha?

niooss-ledger:
> > `parse_lib_version` only supports MAJOR.MINOR versions, not MAJOR.MINOR.PATCH ones. This leads `pkcs11-tool -I` to display a library version which is 0.0, when using tpm2-pkcs11 1.6.0:
> 
> So is this is the issue (how did every other release work?) or is it the shallow clone and getting just a truncated sha?

The is the real issue which is fixed in this PR. On Arch Linux (using 1.6.0), I get "0.0" with:

```text
WARNING: Could not strtoul(6.0): Success
```

The shallow clone from the CI test is just another issue which becomes clear because I added a test case. When I run the tests in containers on my machine, they all pass, including the new one I added.

williamcroberts:
> > > `parse_lib_version` only supports MAJOR.MINOR versions, not MAJOR.MINOR.PATCH ones. This leads `pkcs11-tool -I` to display a library version which is 0.0, when using tpm2-pkcs11 1.6.0:
> > 
> > 
> > So is this is the issue (how did every other release work?) or is it the shallow clone and getting just a truncated sha?
> 
> The is the real issue which is fixed in this PR. On Arch Linux (using 1.6.0), I get "0.0" with:
> 
> ```
> WARNING: Could not strtoul(6.0): Success
> ```
> 
> The shallow clone from the CI test is just another issue which becomes clear because I added a test case. When I run the tests in containers on my machine, they all pass, including the new one I added.

Ahh ok theirs two issues here...
1. Handling just sha's do to shallow clones
2. Handling multiple `.` (which means everything is broken up to this point for versioning via pkcs11)

niooss-ledger:
For the issue "Handling just sha's do to shallow clones", would it be acceptable to fall back to parsing `CHANGELOG.md`?

For example something like this, to search for titles such as `### 1.6.0 - 2021-05-03`:

```patch
diff --git a/configure.ac b/configure.ac
index a7aeaf529c7c..56a278dbe19e 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 AC_INIT([tpm2-pkcs11],
-  [m4_esyscmd_s([git describe --tags --always --dirty])],
+  [m4_esyscmd_s([git describe --tags --always --dirty || (sed -n 's/^### \([0-9.]\+\) - .*/\1/p;' < CHANGELOG.md |head -n1) ])],
   [https://github.com/tpm2-software/tpm2-pkcs11/issues],
   [],
   [https://github.com/tpm2-software/tpm2-pkcs11])
```

williamcroberts:
> For the issue "Handling just sha's do to shallow clones", would it be acceptable to fall back to parsing `CHANGELOG.md`?
> 
> For example something like this, to search for titles such as `### 1.6.0 - 2021-05-03`:
> 
> ```diff
> diff --git a/configure.ac b/configure.ac
> index a7aeaf529c7c..56a278dbe19e 100644
> --- a/configure.ac
> +++ b/configure.ac
> @@ -1,7 +1,7 @@
>  # SPDX-License-Identifier: BSD-2-Clause
>  
>  AC_INIT([tpm2-pkcs11],
> -  [m4_esyscmd_s([git describe --tags --always --dirty])],
> +  [m4_esyscmd_s([git describe --tags --always --dirty || (sed -n 's/^### \([0-9.]\+\) - .*/\1/p;' < CHANGELOG.md |head -n1) ])],
>    [https://github.com/tpm2-software/tpm2-pkcs11/issues],
>    [],
>    [https://github.com/tpm2-software/tpm2-pkcs11])
> ```
> 
> EDIT: well... this does not work because `git describe` does not fail if there is no tag, but this could be fixed with a more complex command which filters on lines containing dots:
> 
> ```diff
> diff --git a/configure.ac b/configure.ac
> index a7aeaf529c7c..6f32c6c3ef76 100644
> --- a/configure.ac
> +++ b/configure.ac
> @@ -1,7 +1,9 @@
>  # SPDX-License-Identifier: BSD-2-Clause
>  
>  AC_INIT([tpm2-pkcs11],
> -  [m4_esyscmd_s([git describe --tags --always --dirty])],
> +  m4_esyscmd_s([
> +    (git describe --tags --always --dirty ; sed -n 's/^### \([0-9.]\+\) - .*/\1/p;' < CHANGELOG.md) | grep '\.' |head -n1
> +  ]),
>    [https://github.com/tpm2-software/tpm2-pkcs11/issues],
>    [],
>    [https://github.com/tpm2-software/tpm2-pkcs11])
> ```
> 
> By the way, I removed the `[...]` around `m4_esyscmd_s` because on Ubuntu 21.04 (autoconf 2.69, automake 1.16.3), `autoreconf` reports:
> 
> ```
> configure.ac:3: warning: AC_INIT: not a literal: m4_esyscmd_s([
> ```

I like the idea, but I specifically wanted 0.0 versions reported by pkcs11 to be known as dev versions, so I don't have cases where 1.6 is not an actual release.

williamcroberts:
I have this fixed and added some unit tests for parsing various version formats over on:
  - https://github.com/tpm2-software/tpm2-pkcs11/pull/689

PR #697, opened 2021-06-11
configure: find the version number from the CHANGELOG if there is no tag

niooss-ledger:
When using git shallow clones (`git clone --depth=1`) or tarball downloads, `git describe --tags --always --dirty` does not give a proper version: it is either a git hash or an empty string.

When such cases occur, fallback to the version which is given in titles of `CHANGELOG.md`.

This fixes an issue discovered as part of https://github.com/tpm2-software/tpm2-pkcs11/pull/693

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 AC_INIT([tpm2-pkcs11],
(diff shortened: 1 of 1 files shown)


williamcroberts:
Always use release tarballs or a clone with enough depth for git describe to work.

PR #715, opened 2021-08-19
tpm2_ptool: import support HMAC

williamcroberts:
Since import doesn't create a specific keyedhash type key (hmac sha256),
we need to support CKK_GENERIC_SECRET. So this adds basic support for
this CKA_KEY_TYPE attribute and allows for more mechanisms for use.


williamcroberts:
This is hung on not having tpm2-tools pointed to master and missing:

```git
commit d5990d9b1d6ddd82935e541c707b8f984c124443
Author: Erik Larsson <who+github@cnackers.org>
Date:   Thu Jul 15 17:03:25 2021 +0200

    tpm2_import: add support for importing HMAC keys and sealed data blobs
    
    Signed-off-by: Erik Larsson <who+github@cnackers.org>
```

PR #717, opened 2021-08-25
add hmac import suppor, fix hmac and support large messages.

williamcroberts:


williamcroberts:
I need to hold this on tpm2-tools:
  - https://github.com/tpm2-software/tpm2-tools/pull/2825

Before releasing this feature, ensure that the CI is on a released tpm2-tools version with that feature.

PR #718, opened 2021-08-31
Fix db up

### LINKED ISSUE #710: DB upgrade function performs out-of-bound array read and might insert mechanism 0
Hello,

Currently in `master` branch, functions `handle_ECDSA_5_to_6` and `handle_AES_5_to_6` of [`src/lib/db.c`](https://github.com/tpm2-software/tpm2-pkcs11/blob/e45e1275d44ae39b9f1badcdc914527a1473be87/src/lib/db.c) might perform out-of-bound array read and introduce a "zero" mechanism in attribute `CKA_ALLOWED_MECHANISMS`.

More precisely, for `handle_ECDSA_5_to_6`, the code does: https://github.com/tpm2-software/tpm2-pkcs11/blob/e45e1275d44ae39b9f1badcdc914527a1473be87/src/lib/db.c#L1855-L1873

As `mech_num` was increased by 3, the `for` loop should use `i < mech_num - 3` instead. Currently, the loop reads `a->pValue[mech_num - 3]` and `a->pValue[mech_num - 2]`, which are right after the higher bound of the array `a->pValue`.

Moreover `handle_ECDSA_5_to_6` continues by adding the 3 new mechanisms at `new_mechs[mech_num - 3]`, `new_mechs[mech_num - 2]` and `new_mechs[mech_num - 1]` and by updating the attribute using `mech_num`: https://github.com/tpm2-software/tpm2-pkcs11/blob/e45e1275d44ae39b9f1badcdc914527a1473be87/src/lib/db.c#L1880-L1884

Therefore if `CKA_ALLOWED_MECHANISMS` already contained `CKM_ECDSA_SHA{256,384,512}`, zeros are inserted in the array. This value maps to `CKM_RSA_PKCS_KEY_PAIR_GEN` (according to <https://github.com/oasis-tcs/pkcs11/blob/0c3ecef6c90dcd8249ba64cc99d50d9f8f73124e/working/identifier_db/raw_ids.db#L342>), which is not what is intended.

For `handle_AES_5_to_6`, the for-loop directly inserts zeros, in: https://github.com/tpm2-software/tpm2-pkcs11/blob/e45e1275d44ae39b9f1badcdc914527a1473be87/src/lib/db.c#L1931-L1947

... because when `if (old_mech == CKM_AES_CBC_PAD || old_mech == CKM_AES_CTR)` happens, the `continue;` increments `i`. This is exactly like what I reported in a previous comment https://github.com/tpm2-software/tpm2-pkcs11/pull/689/files#discussion_r648715640.

Moreover `handle_AES_5_to_6` also has the out-of-bound array access issue, because `mech_num` is incremented by 2 and the `for` loop uses `for (i=0; i < mech_num - 1; i++)`.

I suggest fixing the bounds of the `for` loops, replacing the assignments such as `new_mechs[mech_num - 3] = CKM_ECDSA_SHA256;` with `new_mechs[pos++] = CKM_ECDSA_SHA256;` and then computing `safe_mul(total_bytes, pos, sizeof(CK_MECHANISM_TYPE));`. What do you think?

By the way I did not try the Python code of the 5->6 update path, but the condition for EC keys seem to be inverted, in https://github.com/tpm2-software/tpm2-pkcs11/blob/e45e1275d44ae39b9f1badcdc914527a1473be87/tools/tpm2_pkcs11/db.py#L477-L483

This code seems to be a copy-paste from `_update_on_5`, but there, the condition was for a `continue` statement. As this is now for the "private EC key" case, the code seems wrong (though I have not tested it). Did I miss something?

williamcroberts:


williamcroberts:
@niooss-ledger could you review?

niooss-ledger:
> @niooss-ledger could you review?

Looks good to me :+1: I see that you have already merged this PR, but I guess that it is always good to have someone else take a look at the code, even post-merge.

(Sorry for not having been more responsive, I have been quite busy since my return from holidays)

williamcroberts:
> > @niooss-ledger could you review?
> 
> Looks good to me  I see that you have already merged this PR, but I guess that it is always good to have someone else take a look at the code, even post-merge.
> 
> (Sorry for not having been more responsive, I have been quite busy since my return from holidays)

No worries, if you spot anything file a bug.

PR #721, opened 2021-09-03
Ossl 3 migration

### LINKED ISSUE #686: Support OpenSSL 3.0
Compile against OpenSSL 3.0, fix and verify.

williamcroberts:


src/lib/tpm.c
```diff
@@ -3444,7 +3444,7 @@ static EC_POINT *tpm_pub_to_ossl_pub(EC_GROUP *group, TPM2B_PUBLIC *key) {
         goto out;
     }
 
-    int rc = EC_POINT_set_affine_coordinates_GFp(group,
+    int rc = EC_POINT_set_affine_coordinates(group,
```
gotthardp, line 3447: Please note that EC_POINT_set_affine_coordinates has been added in OpenSSL 1.1.1, see https://www.openssl.org/docs/manmaster/man3/EC_POINT_set_affine_coordinates.html.
If we require OpenSSL >= 1.1.0, we have to live with EC_POINT_set_affine_coordinates_GFp.

akuster:
For what it is worth. Backporting this merge request does compile within the OE env using openssl 3.0.1.

thanks

Issue #700, opened 2021-06-20
tpm2_ptool: ASN1 reparsing might introduces errors


commial:
Hello,

In `pemcert_to_attrs`: https://github.com/tpm2-software/tpm2-pkcs11/blob/8f7b51adeda5b19888e7ad965bd95d0472969cb5/tools/tpm2_pkcs11/utils.py#L264

The code recreates the certificate data from a parsed version, ie: `original -> parsed -> certificate`.

Problem is, **this transformation is not a bijection**. For instance, the ASN1 boolean `True` should be encoded as a not-null value (`1`, `0xFF`, ... are then valid).

As a result, one of the certificate I import is not the same when given back, leading to a broken certificate chain:
```sh
# Parse the certificate
$ openssl asn1parse -in /tmp/cert_only.pem -inform pem
...
  611:d=5  hl=2 l=   3 prim: OBJECT            :X509v3 Key Usage
  616:d=5  hl=2 l=   1 prim: BOOLEAN           :255
  619:d=5  hl=2 l=   4 prim: OCTET STRING      [HEX DUMP]:030205A0
...
# Import it
$ tpm2_ptool addcert --label label --key-id 64616637633462616464393630326431 /tmp/cert_only.pem
# Get it back
$ tpm2_ptool objmod --id 3 --key 17
CKA_VALUE: 3082...
$ python
>>> open("/tmp/cert_out.der", "wb").write(b"3082...")
# Parse it back
$ openssl asn1parse -in /tmp/cert_out.der -inform der
...
  611:d=5  hl=2 l=   3 prim: OBJECT            :X509v3 Key Usage
  616:d=5  hl=2 l=   1 prim: BOOLEAN           :1
  619:d=5  hl=2 l=   4 prim: OCTET STRING      [HEX DUMP]:030205A0
...
# Check the certificate chain
## Original
$ openssl verify -CAfile ~/CA.pem /tmp/cert_only.pem 
/tmp/cert_only.pem: OK
## Outputed
$ openssl x509 -in /tmp/cert_out.der -inform der -out /tmp/cert_out.pem -outform pem
$ openssl verify -CAfile ~/CA.pem /tmp/cert_out.pem
CN = client@WINDOMAIN.LOCAL
error 7 at 0 depth lookup: certificate signature failure
error /tmp/cert_out.pem: verification failed
[...]

Issue #631, opened 2020-11-24
Cannot use certificate imported in tpm


HumanoidPhantom:
Hi!

I have tpm2-pkcs11-1.2.0 installed on Ubuntu 20.04. It is not possible to use certificate, imported in tpm. The following command fails:
```
openssl s_client -engine pkcs11 -keyform engine -key "pkcs11:model=NPCT75x;manufacturer=Nuvoton;serial=0000000000000000;token=test;id=%30%31;object=test-RSA2048-keypair;type=private" -cert "pkcs11:model=NPCT75x;manufacturer=Nuvoton;serial=0000000000000000;token=test;id=%30%31;object=test-RSA2048-keypair;type=cert" -connect example.com:443 -servername example.com -state -debug
engine "pkcs11" set.

Can't open pkcs11:model=NPCT75x;manufacturer=Nuvoton;serial=0000000000000000;token=test;id=%30%31;object=test-RSA2048-keypair;type=cert for reading, No such file or directory
139743449638208:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:69:fopen('pkcs11:model=NPCT75x;manufacturer=Nuvoton;serial=0000000000000000;token=test;id=%30%31;object=test-RSA2048-keypair;type=cert','r')
139743449638208:error:2006D080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:76:
unable to load certificate
```

When I specify path to the certificate on file system, everything works as expected:
```
openssl s_client -engine pkcs11 -keyform engine -key "pkcs11:model=NPCT75x;manufacturer=Nuvoton;serial=0000000000000000;token=test;id=%30%31;object=test-RSA2048-keypair;type=private" -cert test.crt -connect example.com:443 -servername example.com -state -debug
```

The same happens if I test with openconnect, curl or any other application.

Certificate was imported the following way:
```
$ tpm2_ptool addcert --label test --key-label=test-RSA2048-keypair --path=${TPM2_PKCS11_STORE} test.crt
```
tpm2pkcs11-tool shows imported certificate:
```
tpm2pkcs11-tool --list-objects --slot 1
...
Certificate Object; type = X.509 cert
  label:      test-RSA2048-keypair
  subject:    DN: O=Example, CN=Test$
  ID:         3031
```

Can you please advise what might be the problem?

williamcroberts:
[...]

PR #729, opened 2021-09-23
tpm2_ptools: do not re-encode the signed data when importing a certificate

niooss-ledger:
When using `tpm2_ptool addcert`, several users experienced issues because the signed data of the certificate was re-encoded when being added to the database. More precisely, the encoded certificate data is encoded using a BER encoder which encodes booleans using 1 of True (cf. https://github.com/etingof/pyasn1/blob/v0.4.8/pyasn1/codec/ber/encoder.py#L164). But in DER, the encoding of "True" is 0xff, and changing the signed data made the signature of the certificate no longer valid.
    
To fix this issue:
    
- Directly use the result of `pem.readPemFromFile(f)` in attribute `CKA_VALUE`: this is directly the encoded form of the certificate.
- Remove `pyasn1.codec.ber`, as this encoder is no longer used.
- Rename the DER decoder from `decoder` to `derdecoder` and the encoder from `derenc` to `derencoder`, to make the code easier to read.

Also add a test which imports a certificate with boolean values and ensure that the signature is still valid after the import.

While at it, a few other things were modified, which are described in the commit descriptions.

This should fix https://github.com/tpm2-software/tpm2-pkcs11/issues/700 and probably also https://github.com/tpm2-software/tpm2-pkcs11/issues/631

### DIFF
--- a/tools/tpm2_pkcs11/utils.py
+++ b/tools/tpm2_pkcs11/utils.py
@@ -15,9 +15,7 @@ from cryptography.hazmat.primitives.asymmetric import (rsa, padding)
 from cryptography.hazmat.primitives import hashes
 
 from pyasn1_modules import pem, rfc2459
-from pyasn1.codec.der import decoder
-from pyasn1.codec.ber import encoder as berenc
-from pyasn1.codec.der import encoder as derenc
+from pyasn1.codec.der import decoder as derdecoder, encoder as derencoder
 from pyasn1.type import namedtype, tag, univ
 
 from .pkcs11t import *  # noqa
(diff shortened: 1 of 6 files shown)


tools/tpm2_pkcs11/utils.py
```diff
@@ -247,68 +245,64 @@ def asn1_format_ec_point_uncompressed(x, y):
...
-        }
+    # rather than using pycryptography x509 parser, which gives native type access to certificate
+    # fields use pyASN1 to get raw ASN1 encoded values for the fields as the spec requires them
+    with open(certpath, "r") as f:
+        bercert = pem.readPemFromFile(f)
+
+    cert = derdecoder.decode(bercert, asn1Spec=rfc2459.Certificate())[0]
+    c = cert['tbsCertificate']
+
+    # print(cert.prettyPrint())
```
williamcroberts, line 256: drop this....

williamcroberts (commented):
minor nit LGTM

tools/tpm2_pkcs11/utils.py
niooss-ledger: done

Issue #730, opened 2021-09-27
C_Encrypt() doesn't always work with CKM_AES_CBC_PAD


rgerganov:
I have the following code:
```
    rv = pP11Functions->C_EncryptInit(hSession, pMech, hSymKey);
    if(rv != CKR_OK) {
        goto doneEnc;
    }

    // Get the size of the encrypted data.
    rv = pP11Functions->C_Encrypt(hSession, pData, cbData, NULL, pcbEncrypted);
    if(rv != CKR_OK) {
        goto doneEnc;
    }
    printf("Size of encrypted data: %d\n", *pcbEncrypted);

    // Allocate the encypted data buffer.
    *ppEncrypted = (CK_BYTE_PTR)calloc(*pcbEncrypted, 1);
    if(NULL == *ppEncrypted) {
        goto doneEnc;
    }

    // Encrypt the data.
    rv = pP11Functions->C_Encrypt(hSession, pData, cbData, *ppEncrypted, pcbEncrypted);
    if(rv != CKR_OK) {
        goto doneEnc;
    }
```

When I try to encrypt the message `Hello World!` with AES key and `CKM_AES_CBC_PAD` mechanism, I am getting `CKR_BUFFER_TOO_SMALL` error on the second call of `C_Encrypt`. The first call of `C_Encrypt` is successful and it returns the correct size of the encrypted data which is 16. If I do only one call to `C_Encrypt` with the correct buffer size, then it succeeds.

niooss-ledger:
Hello,
I am not familiar with PKCS#11 symmetric encryption use-cases. When reading your code, I am wondering whether you are using correctly the PKCS#11 functions. According to the documentation of `C_EncryptInit` (http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959744):

> After calling C_EncryptInit, the application can either call C_Encrypt to encrypt data in a single part; or call C_EncryptUpdate zero or more times, followed by C_EncryptFinal, to encrypt data in multiple parts.  The encryption operation is active until the application uses a call to C_Encrypt or C_EncryptFinal to actually obtain the final piece of ciphertext.  To process additional data (in single or multiple parts), the application MUST call C_EncryptInit again.

[...]

PR #736, opened 2021-11-04
Fix C_Encrypt() and C_Decrypt() when using CKM_AES_CBC_PAD

rgerganov:
When using pkcs11 it is a common pattern to encrypt data like this:

  CK_ULONG_PTR out_len;
  // find the output length
  C_Encrypt(h, in, in_len, NULL, out_len);
  // allocate buffer with out_len bytes and do the encryption

The actual encryption is done by either a single call to C_Encrypt() or
multiple calls to C_EncryptUpdate(). In both cases we use common
functions which keep internal state needed for CKM_AES_CBC_PAD.
The problem is that C_Encrypt() modifies the state even in the case when
it is used for finding the output length. This causes a subsequent call
to C_Encrypt() for the actual encryption to fail.

The same problem exist for C_Decrypt().

This patch fixes the above problems by clearing the internal state at
the beginning of C_Encrypt/C_Decrypt.

Fixes: #730

### DIFF
--- a/src/lib/encrypt.c
+++ b/src/lib/encrypt.c
@@ -365,8 +365,20 @@ CK_RV decrypt_oneshot_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, CK
 
     bool is_buffer_too_small = false;
     CK_ULONG tmp_len = *data_len;
+    CK_RV rv = CKR_GENERAL_ERROR;
 
-    CK_RV rv = decrypt_update_op(ctx, supplied_opdata, encrypted_data, encrypted_data_len,
+    if (!supplied_opdata) {
+        encrypt_op_data *opdata = NULL;
+        rv = session_ctx_opdata_get(ctx, operation_decrypt, &opdata);
+        if (rv != CKR_OK) {
+            return rv;
+        }
+        if (!opdata->use_sw) {
+            tpm_opdata_reset(opdata->cryptopdata.tpm_opdata);
+        }
+    }
+
+    rv = decrypt_update_op(ctx, supplied_opdata, encrypted_data, encrypted_data_len,
             data, &tmp_len);
     if (rv != CKR_OK && rv != CKR_BUFFER_TOO_SMALL) {
         return rv;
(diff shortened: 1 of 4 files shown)


rgerganov:
I have added an integration test

williamcroberts:
So this is tested in the 5_2 return tests in:
  - https://github.com/tpm2-software/tpm2-pkcs11/blob/master/test/integration/pkcs-crypt.int.c#L425

I think the issue is only in CBC pad and  I think the fix is to add the opdata reset in https://github.com/tpm2-software/tpm2-pkcs11/blob/eea18537f76699174f615625c4d84e39acb93f3b/src/lib/encrypt.c#L317 where we handle the 5.2 style returns.

rgerganov:
I don't think `common_final_op()` is the right place to reset `opdata` because it is also used to finalize multipart encryption and in this case the state needs to be preserved if the buffer is too small. The state needs to be reset only for oneshot operations.

williamcroberts:
> I don't think `common_final_op()` is the right place to reset `opdata` because it is also used to finalize multipart encryption and in this case the state needs to be preserved if the buffer is too small. The state needs to be reset only for oneshot operations.

I wonder if one-shot signs and verifies suffer from the same issue then.....

williamcroberts:
> > I don't think `common_final_op()` is the right place to reset `opdata` because it is also used to finalize multipart encryption and in this case the state needs to be preserved if the buffer is too small. The state needs to be reset only for oneshot operations.
> 
> I wonder if one-shot signs and verifies suffer from the same issue then.....

C_Sign seems to be tested and the context of multi-part or oneshot is passed as a variable to the function. However, C_Verify doesn't seem to be tested for 5.2 returns.

So to keep this matching C_Sign, I added oneshot variables to a common_final_ex routine.... Ill have to write a test for C_Verify and do the same I would imagine.

williamcroberts:
> > > I don't think `common_final_op()` is the right place to reset `opdata` because it is also used to finalize multipart encryption and in this case the state needs to be preserved if the buffer is too small. The state needs to be reset only for oneshot operations.
> > 
> > 
> > I wonder if one-shot signs and verifies suffer from the same issue then.....
> 
> C_Sign seems to be tested and the context of multi-part or oneshot is passed as a variable to the function. However, C_Verify doesn't seem to be tested for 5.2 returns.
> 
> So to keep this matching C_Sign, I added oneshot variables to a common_final_ex routine.... Ill have to write a test for C_Verify and do the same I would imagine.

Nevermind C_Verify doesn't take the ULONG pointer for any size fields....

Also, I added this PR which is based off your work but modifies it to be consistent with C_Sign style:
  - https://github.com/tpm2-software/tpm2-pkcs11/pull/739

williamcroberts:
Close this, fixed #739

PR #739, opened 2021-11-11
When using pkcs11 it is a common pattern to encrypt data like this:

### LINKED ISSUE #730: C_Encrypt() doesn't always work with CKM_AES_CBC_PAD

williamcroberts:
CK_ULONG_PTR out_len;
// find the output length
C_Encrypt(h, in, in_len, NULL, out_len);
// allocate buffer with out_len bytes and do the encryption

The actual encryption is done by either a single call to C_Encrypt() or
multiple calls to C_EncryptUpdate(). In both cases we use common
functions which keep internal state needed for CKM_AES_CBC_PAD.
The problem is that C_Encrypt() modifies the state even in the case when
it is used for finding the output length. This causes a subsequent call
to C_Encrypt() for the actual encryption to fail.

The same problem exist for C_Decrypt().

This patch fixes the above problems by clearing the internal state at
the beginning of C_Encrypt/C_Decrypt.

Fixes: #730

Started-by: Radoslav Gerganov <rgerganov@vmware.com>
Modified-by: William Roberts <william.c.roberts@intel.com>

### DIFF
--- a/src/lib/encrypt.h
+++ b/src/lib/encrypt.h
@@ -49,9 +49,10 @@ static inline CK_RV encrypt_update (session_ctx *ctx, unsigned char *part, unsig
     return encrypt_update_op (ctx, NULL, part, part_len, encrypted_part, encrypted_part_len);
 }
 
-CK_RV encrypt_final_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_encrypted_part, unsigned long *last_encrypted_part_len);
+CK_RV encrypt_final_ex (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_encrypted_part, unsigned long *last_encrypted_part_len, bool is_oneshot);
+
 static inline CK_RV encrypt_final (session_ctx *ctx, unsigned char *last_encrypted_part, unsigned long *last_encrypted_part_len) {
-    return encrypt_final_op (ctx, NULL, last_encrypted_part, last_encrypted_part_len);
+    return encrypt_final_ex (ctx, NULL, last_encrypted_part, last_encrypted_part_len, false);
 }
 
 CK_RV decrypt_init_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, CK_MECHANISM *mechanism, CK_OBJECT_HANDLE key);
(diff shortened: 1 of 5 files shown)


src/lib/encrypt.h
```diff
@@ -49,7 +49,12 @@ static inline CK_RV encrypt_update (session_ctx *ctx, unsigned char *part, unsig
     return encrypt_update_op (ctx, NULL, part, part_len, encrypted_part, encrypted_part_len);
 }
 
-CK_RV encrypt_final_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_encrypted_part, unsigned long *last_encrypted_part_len);
+CK_RV encrypt_final_ex (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_encrypted_part, unsigned long *last_encrypted_part_len, bool is_oneshot);
+
+static inline CK_RV encrypt_final_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_encrypted_part, unsigned long *last_encrypted_part_len) {
```
rgerganov, line 54: Having both `encrypt_final_op()` and `encrypt_final_ex()` is confusing. Remove this function and call `encrypt_final_ex()` directly in `encrypt_final()`

src/lib/encrypt.h
```diff
@@ -64,7 +69,12 @@ static inline CK_RV decrypt_update (session_ctx *ctx, unsigned char *part, unsig
         return decrypt_update_op (ctx, NULL, part, part_len, encrypted_part, encrypted_part_len);
 }
 
-CK_RV decrypt_final_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_part, unsigned long *last_part_len);
+CK_RV decrypt_final_ex(session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_part, unsigned long *last_part_len, bool is_oneshot);
+
+static inline CK_RV decrypt_final_op (session_ctx *ctx, encrypt_op_data *supplied_opdata, unsigned char *last_part, unsigned long *last_part_len) {
```
rgerganov, line 74: Same comment here

src/lib/encrypt.h
williamcroberts: I thought I was matching sign, but sign dropped the op and just calls ex as you mention.

rgerganov (approved):
looks good to me

PR #749, opened 2021-12-22
fix: Esys_GetRandom memory leak

aaronUniontech:
Signed-off-by: 张亚 <zhangya@uniontech.com>


src/lib/tpm.c
```diff
@@ -793,19 +793,21 @@ bool tpm_getrandom(tpm_ctx *ctx, BYTE *data, size_t size) {
             &rand_bytes);
         if (rval != TSS2_RC_SUCCESS) {
             LOGE("Esys_GetRandom: %s:", Tss2_RC_Decode(rval));
+            free(rand_bytes);
```
williamcroberts, line 796: While we are at it, use: `Esys_Free()`.

PR #764, opened 2022-04-01
Enforce lock path

vjardin:
When a system is not reliable, the lock file should not be local on
the persistent filesystem. Let's enforce it into a tpmfs.


docs/INITIALIZING.md
```diff
@@ -51,6 +51,12 @@ The store itself defaults to `$HOME/.tpm2_pkcs11` unless specified via the envir
 
   Their is no requirement to use the simulator and abrmd, this is all configuration dependent.
 
+**LOCKING**
+
+When the SQL database is on the disk, the lock is set within the same folder than the SQL file.
+It can lead to some issues if the lock is not released (system crash, reboot), mostly on embedded
+systems. Another folder, for instance a tmpfs one, can be enforced using the env `PKCS11_SQL_LOCK=/var/run/pkcs11_sql_locks`.
```
williamcroberts, line 58: What happens if their are N instances of the library running? We would need N distinct lock files. I am ok with moving it to a tmpfs by default and not needing an env variable for it.

vjardin: Same as using the SQL file: if the instances share the same SQL file so they should share the same lock path.

vjardin: Please, @williamcroberts , are you maintainer ? If so, can you unlock the workflow ?

williamcroberts: > Same as using the SQL file: if the instances share the same SQL file so they should share the same lock path.

So `take_lock()` takes in the store path with db filename, like `/home/bill/mystore/tpm2_pkcs11.sqlite3` and appends `.lock`, so you end up with a `/home/bill/mystore/tpm2_pkcs11.sqlite3.lock`, one lock file per store. Stores are controlled by default searching and an env variable. But in no way could you map a store to a different lock file.

In this case, its possible to get us to a state where you could have store X, Y and Z in use but have them all pointed to one lock file or have store X in use and point it to two different lock files.

You could probably swap all path separators `/` with underscores `_` and then create a file, using my previous full example of `/var/run/pkcs11_locks/home_bill_mystore_tpm2_pkcs11.sqlite3.lock`. Then you would have the same mapping of 1 lock to 1 store.

williamcroberts: > Please, @williamcroberts , are you maintainer ? If so, can you unlock the workflow ?

Sure, but lets get those changes in and ill approve it.

williamcroberts (changes requested):

docs/INITIALIZING.md
vjardin: ok, got it, you are right.
See the new update 59404ca0db6d646532e602fcb30bf5b456239918 that does the substitution of '/' by '_'/

src/lib/db.c
```diff
@@ -2324,7 +2324,26 @@ static CK_RV db_create(char *path, size_t len) {
...
+         * path's '/' will be substituted by '_'.
+         */
+        char path_alt[PATH_MAX];
+
+        strncpy(path_alt, path, PATH_MAX);
+        for (int i = 0; path_alt[i] && (i < PATH_MAX); i++) {
+            if (path_alt[i] == '/')
+                path_alt[i] = '_';
+        }
+        l = snprintf(lockpath, PATH_MAX, "%s%s%s", env_lock, path_alt, ".lock");
```
williamcroberts, line 2343: In your example, you set:
```bash
PKCS11_SQL_LOCK=/var/run/pkcs11_sql_locks`.
```
In the code you do:
```C
snprintf(lockpath, PATH_MAX, "%s%s%s", env_lock, path_alt, ".lock");
/var/run/pkcs11_sql_lockspath_to_store.lock
```
Did you intend for that or to be part of some sub directory? Ie enforce that the last character in env_lock is PATH separator (/)? I would imagine we would want some directory path that we store all the locks under.

williamcroberts: It would also be nice to not use another large stack buffer. Perhaps we just manually build out the string into the lockpath buffer and avoid using strncpy and snprintf. Here is some pseudo-code. 

```C
l = strncpy(lockpath, env_lock, PATH_MAX);
// check for overflow

// check for path sep (/) on env_lock and add it if missing

// update current offset

// for each character in path
while(i < len(path)) {
   lockpath[offset++] = path[i] == '/' ? '_' : path[i];
   if (offset >= PATH_MAX) {
      fail
   }
   i++;
}

if PATH_MAX - offset != space for ".lock" and NULL byte
   fail
```

src/lib/db.c
```diff
@@ -2324,7 +2324,26 @@ static CK_RV db_create(char *path, size_t len) {
...
+    unsigned l;
+
+    char *env_lock = getenv("PKCS11_SQL_LOCK");
+
+    if (env_lock) {
+        /*
+         * lock file shall be "PKCS11_SQL_LOCK" + path + ".lock", but
+         * path's '/' will be substituted by '_'.
+         */
+        char path_alt[PATH_MAX];
```
williamcroberts, line 2336: Let me be clear here, im either looking for a calloc or to use lockpath buffer in place....

williamcroberts (changes requested):

src/lib/db.c
vjardin: Frankly, I do not understand the benefits. A `path_alt[PATH_MAX]` seems to be a large buffer, but it won't hurt ; they are just few bytes. Your proposal can work, but it makes the code more complex to follow/read.

I agree it is a matter of point of view, but nowadays, I feel it becomes a descent practice to have large stack buffers, as far as it remains under control.

src/lib/db.c
vjardin: thanks @williamcroberts , I do understand your proposals, but please see my comment: I feel we could keep it simple.

src/lib/db.c
williamcroberts: > Frankly, I do not understand the benefits. A `path_alt[PATH_MAX]` seems to be a large buffer, but it won't hurt ; they are just few bytes. Your proposal can work, but it makes the code more complex to follow/read.
> 
> I agree it is a matter of point of view, but nowadays, I feel it becomes a descent practice to have large stack buffers, as far as it remains under control.

A few things, we've had to do some cleanups on tcti project around this:
  - https://github.com/tpm2-software/tpm2-tss/pull/2319
  - https://github.com/tpm2-software/tpm2-tss/pull/2321
  
so id like to not introduce these to remove later.

Next if a buffer overflow occurs, it's on the stack, which means that the control structures for branching are there as well (call stack), so an overflow on the stack is easier to get control of the machine than it is through buffer overflows (possible but maybe/usually harder).

Lastly a calloc copy and free isn't too bad. We can just do that.

williamcroberts: It's actually not a few bytes, its 4096 bytes on linux, which is a whole page.

docs/INITIALIZING.md
vjardin: done

src/lib/db.c
vjardin: done with the commit 937af5884c5ea2eebdc3bb46ee58c19930abcf79

FYI, on Linux Ubuntu 20.04, it is only 128 bytes:
```
  printf("PATH_MAX=%d\n", PATH_MAX);
 ...
PATH_MAX=128
```

src/lib/db.c
vjardin: done with the commit 937af5884c5ea2eebdc3bb46ee58c19930abcf79

williamcroberts (approved):
This looks ok, I'll write a fuzz test to make sure we don't have any string handling issues when I merge it. I just let the CI run, so let's see if it comes back with any issues.

williamcroberts:
Tests and a minor tweak on: https://github.com/tpm2-software/tpm2-pkcs11/pull/773

PR #786, opened 2022-08-24
Modifications for not reading objects into token at C_Initialize, but everytime C_FindObjectInit is called

pepipox:
In order to keep the list in C_FindObjects up to date in case the DB has been updated externally (by tpm2_ptool), we now delete unconditionally the list of objects and read from the database everytime C_FindObjectInit is called.

### DIFF
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -225,7 +225,7 @@ error:
     return rc;
 }
 
-int init_tobjects(token *tok) {
+WEAK DEBUG_VISIBILITY int init_tobjects(token *tok) {
     return __real_init_tobjects(tok);
 }
 
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -225,7 +225,7 @@ error:
     return rc;
 }
 
-WEAK DEBUG_VISIBILITY int init_tobjects(token *tok) {
+int init_tobjects(token *tok) {
     return __real_init_tobjects(tok);
 }
 
@@ -633,11 +633,6 @@ CK_RV db_get_tokens(token *tok, size_t *len) {
         }
 
         rc = init_sealobjects(t->id, &t->esysdb.sealobject);
-        if (rc != SQLITE_OK) {
-            goto error;
-        }
-
-        rc = init_tobjects(t);
         if (rc != SQLITE_OK) {
             goto error;
         }
--- a/src/lib/db.h
+++ b/src/lib/db.h
@@ -72,7 +72,7 @@ CK_RV db_update_token_config(token *tok);
 
 CK_RV db_update_tobject_attrs(unsigned id, attr_list *attrs);
 
-int init_tobjects(token *tok);
+WEAK DEBUG_VISIBILITY int init_tobjects(token *tok);
 
 /* Debug testing */
 #ifdef TESTING
--- a/src/lib/db.h
+++ b/src/lib/db.h
@@ -72,7 +72,7 @@ CK_RV db_update_token_config(token *tok);
 
 CK_RV db_update_tobject_attrs(unsigned id, attr_list *attrs);
 
-WEAK DEBUG_VISIBILITY int init_tobjects(token *tok);
+WEAK int init_tobjects(token *tok);
 
 /* Debug testing */
 #ifdef TESTING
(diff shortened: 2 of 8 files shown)


src/lib/db.c
```diff
@@ -225,7 +225,7 @@ DEBUG_VISIBILITY int __real_init_tobjects(token *tok) {
     return rc;
 }
 
-WEAK DEBUG_VISIBILITY int init_tobjects(token *tok) {
+int init_tobjects(token *tok) {
```
williamcroberts, line 228: This might still need week for override in test/unit/test_db.c

williamcroberts:
Please sign off your commit, the DCO bot failing you has instructions how

src/lib/object.c
```diff
@@ -335,6 +336,13 @@ CK_RV object_find_init(session_ctx *ctx, CK_ATTRIBUTE_PTR templ, CK_ULONG count)
     token *tok = session_ctx_get_token(ctx);
     assert(tok);
 
+    session_ctx_delete_tobject_list(ctx);
+
+    rc = init_tobjects(tok);
```
williamcroberts, line 341: We can probably drop the initial init call to make token initialization quicker:
in `src/lib/db.c` line 640 delete the following statement:
```C
        rc = init_tobjects(t);
        if (rc != SQLITE_OK) {
            goto error;
        }
```

williamcroberts (changes requested):
Changes:
- DCO Signoff
- Maybe need to keep WEAK on init_tobjects
- Drop tobject list initialization in token initialization code

src/lib/object.c
pepipox: Hi William. that call is already dropped. Objects are only loaded in C_FinitObjectInit

williamcroberts: > Hi William. that call in src/lib/db.c is already dropped. Objects are only loaded in C_FinitObjectInit

Oh I missed that.

src/lib/db.h
```diff
@@ -72,6 +72,8 @@ CK_RV db_update_token_config(token *tok);
 
 CK_RV db_update_tobject_attrs(unsigned id, attr_list *attrs);
 
+WEAK DEBUG_VISIBILITY int init_tobjects(token *tok);
```
williamcroberts, line 75: Drop DEBUG_VISIBILITY

src/lib/object.c
williamcroberts, line 341:
```C
int rc = init_tobjects(tok);
```

src/lib/object.c
```diff
@@ -310,6 +310,7 @@ static CK_RV do_match_set(tobject_match_list *match_cur, tobject *tobj) {
 }
 
 CK_RV object_find_init(session_ctx *ctx, CK_ATTRIBUTE_PTR templ, CK_ULONG count) {
+    int rc;
```
williamcroberts, line 313: Just declare it on line 341 with the first use.

williamcroberts (changes requested):
2 minor nits

williamcroberts:
I'll need to fix the CI

src/lib/object.c
pepipox: Done

src/lib/db.h
pepipox: Done

pepipox:
Thanks for all the help

williamcroberts:
> Thanks for all the help

Ok I fixed it in: https://github.com/tpm2-software/tpm2-pkcs11/pull/787

Can you rebase on top of current master and re-push your changes?

src/lib/object.c
```diff
@@ -310,7 +310,7 @@ static CK_RV do_match_set(tobject_match_list *match_cur, tobject *tobj) {
 }
 
 CK_RV object_find_init(session_ctx *ctx, CK_ATTRIBUTE_PTR templ, CK_ULONG count) {
-
+
```
williamcroberts, line 313: superfluous whitespace change, drop

pepipox:
Hi @williamcroberts 
I had rebased as you requested, could you please check the changes? thank you and sorry for the delay, very busy at the office these days.

williamcroberts (changes requested):
minor nit

PR #798, opened 2022-11-22
Set use_createloaded

nickbedbury:
I noticed that `tpm->use_createloaded` wasn't being set anywhere. 

I think this might be close to the correct functionality, but open to suggestions!


williamcroberts:
> I noticed that `tpm->use_createloaded` wasn't being set anywhere.
> 
> I think this might be close to the correct functionality, but open to suggestions!

Looks good, please sign off your commit. You can do `git commit --ammend -s` and force push your branch with `git push -f`.

williamcroberts (approved):

Issue #805, opened 2023-01-06
32/64 bits systems - long encoding SQL: inconsitencies


vjardin:
From `emitter.c` the snprintf("%lu") will have a different behavior if it is run using 32 bits or 64 bits. So, the content of the SQL file will be different. It means, for instance, if the content of the database is build/populated using a 32 compiled version of tpm2-pkcs11, then another set of applications running 64 bits on the same system won't be able to use properly the contents.

It seems to be due to the 3 following print formats:
```
85:        snprintf(strkey, sizeof(strkey), "%lu", k);
128:                snprintf(keyvaluebuf, sizeof(keyvaluebuf), "%lu", *v);
176:                snprintf(keyvalue, sizeof(keyvalue), "%lu", v);
```

what if we replace them with snprintf("xxx %"PRIu64" xxx") ? so the encoding would remain consistent on any execution environment ?

vjardin:
notes:
```
src/pkcs11.h:#define CKA_ALLOWED_MECHANISMS		(CKF_ARRAY_ATTRIBUTE | 0x600UL)
```

it means:
```
CKA_ALLOWED_MECHANISMS=0x40000600
```

so, doing `select id, tokid, attrs from tobjects where id = 1;`
leads to the following attrs:
```
---
!!map {
   ...
  ? !!int "1073743360"  -> 0x40000600
  : !!str "030000000900000001000000060000004000000041000000420000000d0000000e000000430000004400000045000000",
   ...
}
```

vjardin:
Following some deeper analysis, it is even more weird, there are 2 sets of data into the sql files for the CKA_ALLOWED_MECHANISMS:

bad on 32 bits/64 bits:
```
  ? !!int "1073743360"
  : !!str "030000000900000001000000060000004000000041000000420000000d0000000e000000430000004400000045000000",
```

good:
```
   ? !!int "1073743360"
  : !!seq [
    !!int "3",
    !!int "9",
    !!int "1",
    !!int "64",
    !!int "65",
    !!int "66",
    !!int "13",
    !!int "14",
    !!int "67",
    !!int "68",
  ],
```

I do not understand how the bad case entries have been added into the sqlite file !!!

So, maybe the previous statement of trying to convert 32/64 bits is not the problem. The root case, is how/why was a string used for 40000600 / 1073743360 / CKA_ALLOWED_MECHANISMS

vjardin:
[...]

PR #806, opened 2023-01-07
avoid %lu since size is not consistent

vjardin:
Let' use PRIu64 formatting so the length will be the same on both 64 bits and 32 bits system.

### DIFF
--- a/src/lib/emitter.c
+++ b/src/lib/emitter.c
@@ -2,6 +2,7 @@
 #include "config.h"
 #include <assert.h>
 #include <stdlib.h>
+#include <inttypes.h>
 
 #include <yaml.h>
 
@@ -82,7 +83,7 @@ WEAK char *emit_attributes_to_string(attr_list *attrs) {
 
         char strkey[64] = { 0 };
         CK_ATTRIBUTE_TYPE k = a->type;
-        snprintf(strkey, sizeof(strkey), "%lu", k);
+        snprintf(strkey, sizeof(strkey), "%"PRIu64"", k);
 
         int key = yaml_document_add_scalar(&doc, (yaml_char_t *)YAML_INT_TAG,
              (yaml_char_t *)strkey, -1, YAML_ANY_SCALAR_STYLE);
@@ -125,7 +126,7 @@ WEAK char *emit_attributes_to_string(attr_list *attrs) {
             const char *keyvalue = NULL;
             if (type == TYPE_BYTE_INT) {
                 CK_ULONG_PTR v = (CK_ULONG_PTR)a->pValue;
-                snprintf(keyvaluebuf, sizeof(keyvaluebuf), "%lu", *v);
+                snprintf(keyvaluebuf, sizeof(keyvaluebuf), "%"PRIu64"", *v);
                 keyvalue = keyvaluebuf;
             } else if (type == TYPE_BYTE_BOOL) {
                 CK_BBOOL *v = (CK_BBOOL *)a->pValue;
@@ -173,7 +174,7 @@ WEAK char *emit_attributes_to_string(attr_list *attrs) {
                 CK_ULONG v = ((CK_ULONG_PTR)a->pValue)[j];
 
                 char keyvalue[64] = { 0 };
-                snprintf(keyvalue, sizeof(keyvalue), "%lu", v);
+                snprintf(keyvalue, sizeof(keyvalue), "%"PRIu64"", v);
 
                 /* create a scalar */
                 int seqscalar = yaml_document_add_scalar(&doc, (yaml_char_t *)YAML_INT_TAG,


vjardin:
fix of https://github.com/tpm2-software/tpm2-pkcs11/issues/805

vjardin:
Please, @williamcroberts , what does the CI keeps failing ? 
For instance:
```
docker: Error response from daemon: Head "https://ghcr.io/v2/tpm2-software/fedora-30/manifests/latest": unauthorized.
```
I do not understand how to work around it.

vjardinMisc:
Please, @williamcroberts , can you check the following for fixing the CI ?
  https://github.com/nektos/act/issues/1248#issuecomment-1178309705

or can you check your PAT (Personal Access Token)? It may have expired.
I believe, it is yours, not mine ; but it is unclear.

williamcroberts:
> Please, @williamcroberts , can you check the following for fixing the CI ? [nektos/act#1248 (comment)](https://github.com/nektos/act/issues/1248#issuecomment-1178309705)
> 
> or can you check your PAT (Personal Access Token)? It may have expired. I believe, it is yours, not mine ; but it is unclear.

I cleaned up our CI images and thought fedora-30 was unused, I restored the image.

src/lib/emitter.c
```diff
@@ -82,7 +83,7 @@ WEAK char *emit_attributes_to_string(attr_list *attrs) {
 
         char strkey[64] = { 0 };
         CK_ATTRIBUTE_TYPE k = a->type;
-        snprintf(strkey, sizeof(strkey), "%lu", k);
+        snprintf(strkey, sizeof(strkey), "%"PRIu64"", k);
```
williamcroberts, line 86: on 32 bit systems this causes a warning:
```
a.c: In function 'main':
a.c:6:9: warning: format '%llu' expects argument of type 'long long unsigned int', but argument 2 has type 'long unsigned int' [-Wformat=]
    6 |  printf("%"PRIu64"\n", x);
      |         ^~~            ~
```

williamcroberts (changes requested):
The type is unsigned long, so converting to and from the type should use the `%lu` format string. Their will be no loss of precision because the data type is of that precision and will be written out and read in from the YAML in the precision that is supported.

```C
typedef unsigned long ck_attribute_type_t;
#define ck_attribute_type_t CK_ATTRIBUTE_TYPE
```

Nack, this doesn't fix anything, this actually introduces issues.

vjardin:
> The type is unsigned long, so converting to and from the type should use the `%lu` format string. Their will be no loss of precision because the data type is of that precision and will be written out and read in from the YAML in the precision that is supported.
> 
> ```c
> typedef unsigned long ck_attribute_type_t;
> #define ck_attribute_type_t CK_ATTRIBUTE_TYPE
> ```
> 
> Nack, this doesn't fix anything, this actually introduces issues.

yes, but if you write 32bits and a 64 bits application read it you get some issues.

You write:
  "030000000900000001000000060000004000000041000000420000000d0000000e000000430000004400000045000000"
but you'd like to read:
  "03000000000000000900000000000000010000000000000006000000000000004000000000000000410000000000000042000000000000000d000000000000000e00000000000000430000000000000044000000000000004500000000000000"

src/lib/emitter.c
vjardin: Back to the basics, we could enforce a cast:
```
snprintf(strkey, sizeof(strkey), "%"PRIu64"", (u_int64_t)k);
```
but first, let's make sure we agree on the problem.

vjardin:
Per the comments https://github.com/tpm2-software/tpm2-pkcs11/issues/805#issuecomment-1374878320 ; it is another problem to be fixed ! yaml seq vs yaml string ?! for a CKA_ALLOWED_MECHANISMS

williamcroberts:
> Per the comments [#805 (comment)](https://github.com/tpm2-software/tpm2-pkcs11/issues/805#issuecomment-1374878320) ; it is another problem to be fixed ! yaml seq vs yaml string ?! for a CKA_ALLOWED_MECHANISMS

CKA_ALLOWED_MECHANISMS should be a sequence:

```bash
tpm2_ptool objmod --key=CKA_ALLOWED_MECHANISMS --id=12 --path=$HOME/tmp
CKA_ALLOWED_MECHANISMS:
- 4161
- 4162
- 4164
- 4165
- 4166
```
Are you getting a YAML file generated where that's not the case? If so, how can I reproduce it and whats the contents of the whole YAML data? You can use objmod to dump it.

```bash
tpm2_ptool objmod --id=12 --path=$HOME/tmp
0: 3
1: true
2: true
3: '656330'
256: 3
257: ''
258: '65656636653364613638326362633862'
259: true
261: true
<snip>
```

PR #807, opened 2023-01-07
use github_token for ghcr login

vjardin:
Fix the error:
  docker: Error response from daemon: Head "https://ghcr.io/v2/tpm2-software/fedora-30/manifests/latest": unauthorized.


vjardin:
@williamcroberts >  does it worth keeping this pull request, or should we drop/cancel it ?

williamcroberts:
Drop it, we don't need to use the secrets for this

PR #808, opened 2023-01-09
CKA_ALLOWED_MECHANISMS - it should be a int_seq

vjardin:
memtype was not properly set.

### DIFF
--- a/src/lib/db.c
+++ b/src/lib/db.c
@@ -32,6 +32,7 @@
 #include "tpm.h"
 #include "twist.h"
 #include "utils.h"
+#include "typed_memory.h"
 
 #include <openssl/evp.h>
 
@@ -39,7 +40,7 @@
 #define TPM2_PKCS11_STORE_DIR "/etc/tpm2_pkcs11"
 #endif
 
-#define DB_VERSION 7
+#define DB_VERSION 8
 
 #define goto_oom(x, l) if (!x) { LOGE("oom"); goto l; }
 #define goto_error(x, l) if (x) { goto l; }
@@ -2131,6 +2132,69 @@ error:
     return rv;
 }
 
+static CK_RV dbup_handler_from_7_to_8(sqlite3 *updb) {
+
+    /*
+     * Between version 7 and 8 of the DB the following changes need to be made:
+     *
+     * Table tobjects:
+     *
+     * The YAML attributes for a int sequence has to be a yaml sequence.
+     */
+
+
+    CK_RV rv = CKR_GENERAL_ERROR;
+    sqlite3_stmt *stmt = NULL;
+
+    int rc = sqlite3_prepare_v2(updb, "SELECT * from tobjects", -1, &stmt, 0);
+    if (rc != SQLITE_OK) {
+        LOGE("Failed to fetch data: %s", sqlite3_errmsg(updb));
+        goto error;
+    }
+
+    rc = sqlite3_step(stmt);
+    if (rc == SQLITE_DONE) {
+        goto out;
+    } else if (rc != SQLITE_ROW) {
+        LOGE("Failed to step: %s", sqlite3_errmsg(updb));
+        goto error;
+    }
+
+    while (rc == SQLITE_ROW) {
+        tobject *tobj = db_tobject_new(stmt);
+        if (!tobj) {
+            LOGE("Could not process tobjects for upgrade");
+            goto error;
+        }
+
+        /* for each tobject */
+        CK_ATTRIBUTE_PTR a = attr_get_attribute_by_type(tobj->attrs, CKA_ALLOWED_MECHANISMS);
+        CK_BYTE type = type_from_ptr(a->pValue, a->ulValueLen);
+        if (type != TYPE_BYTE_INT_SEQ) {
+            rv = _db_update_tobject_attrs(updb, tobj->id, tobj->attrs);
+        }
+
+        tobject_free(tobj);
+        if (rv != CKR_OK) {
+            goto error;
+        }
+
+        rc = sqlite3_step(stmt);
+        if (rc != SQLITE_ROW && rc != SQLITE_DONE) {
+            LOGE("Failed to fetch data: %s\n", sqlite3_errmsg(updb));
+            goto error;
+        }
+    }
+
+out:
+    rv = CKR_OK;
+
+error:
+    sqlite3_finalize(stmt);
+    return rv;
+}
+
+
 static CK_RV db_backup(sqlite3 *db, const char *dbpath, sqlite3 **updb, char **copypath) {
 
     CK_RV rv = CKR_GENERAL_ERROR;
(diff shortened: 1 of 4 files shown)


vjardin:
better option in order to fix https://github.com/tpm2-software/tpm2-pkcs11/issues/805

vjardin:
then, side question @williamcroberts > since the YAML content of the SQL database has to be cleaned/updated in order to convert the str: to a seq: for a production system, what would be the best option in order to do so ?

williamcroberts (approved):

williamcroberts:
> then, side question @williamcroberts > since the YAML content of the SQL database has to be cleaned/updated in order to convert the str: to a seq: for a production system, what would be the best option in order to do so ?

We will need to bump the database number and add a handler to fix this.

An example of a handler is:
  - https://github.com/tpm2-software/tpm2-pkcs11/blob/master/src/lib/db.c#L1586

This handler would look at the YAML contents for each object and check its CKA_ALLOWED_MECHASMS is a sequence and not a hex string, if it is, it'll need to parse it apart and update it to the sequence list by calling the YAML emitter again.

Then you add the handler to this list at the end:
  - https://github.com/tpm2-software/tpm2-pkcs11/blob/master/src/lib/db.c#L2201

Then you bump the db version number:
  - https://github.com/tpm2-software/tpm2-pkcs11/blob/master/src/lib/db.c#L42

vjardin:
thanks @williamcroberts , it'll be checked.

And regarding the current fix, does it look good ?

vjardin:
Regarding the marshaling and the rewrite of the yaml content, I am not an expert with the libyaml, do you have a tip for doing it in a smart way ?

williamcroberts:
> Regarding the marshaling and the rewrite of the yaml content, I am not an expert with the libyaml, do you have a tip for doing it in a smart way ?

The attributes should be right in the "tobject", so I would just call the emitter and replace their contents. You could probably call `_db_update_tobject_attrs(db, tobj->id, tobs->attrs)`;

To detect if they need updating, I would call:
```C
for each tobject:
    CK_ATTRIBUTE_PTR a = attr_get_attribute_by_type(tobj->attrs, CKA_ALLOWED_MECHANISMS);
    CK_BYTE type = type_from_ptr(a->pValue, a->ulValueLen);
    if (type != TYPE_BYTE_INT_SEQ) {
        _db_update_tobject_attrs(db, tobj->id, tobs->attrs);
    }
```

vjardin:
@williamcroberts > is it the proper marshalling ?
  https://github.com/tpm2-software/tpm2-pkcs11/pull/808/commits/ef1547b62ce70b394b5a74aa15567abf9c445463
  
please, do not merge it yet, it is a draft code for sharing my understanding based on your comments.

vjardin:
I do not understand how to get the sonar report from these logs:
```
scan-build: Run 'scan-view /tmp/scan-build-2023-01-10-161330-53449-1' to examine bug reports.
Error: Process completed with exit code 1.
```
 ??

williamcroberts:
> @williamcroberts > is it the proper marshalling ? [ef1547b](https://github.com/tpm2-software/tpm2-pkcs11/commit/ef1547b62ce70b394b5a74aa15567abf9c445463)
> 
> please, do not merge it yet, it is a draft code for sharing my understanding based on your comments.

looks about right, i'd have to dig into why its failing to give you better feedback.

vjardin:
> looks about right, i'd have to dig into why its failing to give you better feedback.

Can we run the test with more verbosity in order to understand the root cause of the failures ?

williamcroberts:
> > looks about right, i'd have to dig into why its failing to give you better feedback.
> 
> Can we run the test with more verbosity in order to understand the root cause of the failures ?

It tells you if select the failure tab:
```
+ tpm2_ptool addcert --label=label --key-label=ec1 --path=/tmp/tpm_simulator_BBsIAP /tmp/tpm_simulator_BBsIAP/cert.pem.ec1
DB Upgrade failed: "DB Version exceeds library version: 8 > 7"
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
```

I forgot the python code also needs the same logic added to it.

See the logs
https://github.com/tpm2-software/tpm2-pkcs11/actions/runs/3885289344/jobs/6628978268#step:4:365

vjardin:
> I forgot the python code also needs the same logic added to it.

thanks for the tip, python version push'd, so far so good for the CI ;)

tools/tpm2_pkcs11/db.py
```diff
@@ -532,6 +532,23 @@ def _update_on_7(self, dbbakcon):
...
+
+        The YAML attributes for a int sequence has to be a yaml sequence.
+        '''
+        c = dbbakcon.cursor()
+
+        c.execute('SELECT * from tobjects')
+        tobjs = c.fetchall()
+
+        for t in tobjs:
+            attrs = yaml.safe_load(io.StringIO(t['attrs']))
```
williamcroberts, line 549: does this actually work, if you had a V7 database? As t['attrs'] is either a list or a string, shouldn't the check be:
1. if string, break it apart based on sizeof the data type? You could use something `ctypes.sizeof(c_ulong)`, then take that result multiply it by two and split the hex strings on that and create a list, then convert that list to yaml update the attributes and then update the DB

williamcroberts (changes requested):

tools/tpm2_pkcs11/db.py
vjardin: right, it was wrong ! hold on.

vjardin: @williamcroberts , is https://github.com/tpm2-software/tpm2-pkcs11/commit/f15726cf91de0d5f4612a957e981981b020f1835 what you'd like ?

williamcroberts: > @williamcroberts , is [f15726c](https://github.com/tpm2-software/tpm2-pkcs11/commit/f15726cf91de0d5f4612a957e981981b020f1835) what you'd like ?

Yeah that looks right, just some minor nits on that patch

vjardinMisc: Which ones ? Which modifications would you like?

williamcroberts (approved):

vjardinMisc:
Hi @williamcroberts , I think it is all done. If any other requests, feel free.

williamcroberts:
opensuse-leap CI is having issues, but I think it's caused from an update to the images and not this patch, so merging.

src/lib/db.c
```diff
@@ -2131,6 +2132,69 @@ static CK_RV dbup_handler_from_6_to_7(sqlite3 *updb) {
...
+        tobject *tobj = db_tobject_new(stmt);
+        if (!tobj) {
+            LOGE("Could not process tobjects for upgrade");
+            goto error;
+        }
+
+        /* for each tobject */
+        CK_ATTRIBUTE_PTR a = attr_get_attribute_by_type(tobj->attrs, CKA_ALLOWED_MECHANISMS);
+        CK_BYTE type = type_from_ptr(a->pValue, a->ulValueLen);
+        if (type != TYPE_BYTE_INT_SEQ) {
```
williamcroberts, line 2173: I think we needed to change the data type here to TYPE_BYTE_INT_SEQ so the emitter updates everything, so I think this didn't actually do anything.

williamcroberts: I have a PR for this fix: https://github.com/tpm2-software/tpm2-pkcs11/pull/924 which also includes the python code fix from the discussion in #805

PR #810, opened 2023-02-01
Minor FAPI cleanups

u1f35c:
This PR contains 2 minor FAPI cleanups

- Fix the help text to show enable-fapi rather than with-fapi to enable/disable the support
- Don't spit out a warning message about not being able to fetch from FAPI when it's not compiled in and it's not the selected backend

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -116,7 +116,7 @@ AS_IF([test "x$esapi_sf" = "xyes"],
 
 AC_ARG_ENABLE(
   [fapi],
-  [AS_HELP_STRING([--with-fapi],
+  [AS_HELP_STRING([--enable-fapi],
     [enable or disable the fapi backend. Default is "auto" to autodetect])],
     [enable_fapi=$enableval],
     [enable_fapi=auto])
--- a/src/lib/backend.c
+++ b/src/lib/backend.c
@@ -53,7 +53,9 @@ CK_RV backend_init(void) {
             LOGE(msg);
             return rv;
         }
-        LOGW(msg);
+        if (rv != CKR_FUNCTION_NOT_SUPPORTED) {
+            LOGW(msg);
+        }
     } else {
         fapi_init = true;
     }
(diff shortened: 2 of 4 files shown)


configure.ac
```diff
@@ -116,7 +116,7 @@ AS_IF([test "x$esapi_sf" = "xyes"],
 
 AC_ARG_ENABLE(
   [fapi],
-  [AS_HELP_STRING([--with-fapi],
+  [AS_HELP_STRING([--enable-fapi],
```
williamcroberts, line 119: Changing this breaks those building packages for very little gain. While I prefer enable-fapi to with-fapi, I don't want to break downstream packagers for something this trivial.

src/lib/backend.c
```diff
@@ -53,7 +53,9 @@ CK_RV backend_init(void) {
             LOGE(msg);
             return rv;
         }
-        LOGW(msg);
+        if (rv != CKR_FUNCTION_NOT_SUPPORTED) {
+            LOGW(msg);
+        }
```
williamcroberts, line 58: Ill take this change

src/lib/backend_fapi.c
```diff
@@ -977,7 +977,8 @@ CK_RV backend_fapi_token_changeauth(token *tok, bool user, twist toldpin, twist
 
 CK_RV backend_fapi_init(void) {
 
-	return CKR_OK;
+	LOGV("FAPI not enabled, failing init");
+	return CKR_FUNCTION_NOT_SUPPORTED;
```
williamcroberts, line 981: ill take this change

williamcroberts (changes requested):

configure.ac
u1f35c: This only changes the help text from configure to match the actual parameter that's supported. `docs/BUILDING.md` already references `--enable-fapi`. I've got no objections to using `--with-fapi` but that's not what `AC_ARG_ENABLE` does, it should be `AC_ARG_WITH` instead which *would* change the behaviour for downstream packagers.

(I only noticed this when trying to explicitly disable FAPI in the Debian package and finding the help text was wrong.)

williamcroberts: This was actually right... See https://github.com/tpm2-software/tpm2-pkcs11/pull/824. I need to revert this one patch

PR #816, opened 2023-02-16
Fix failing database upgrade

tynsh:
Fix double conversion to int, causing a failing database upgrade.


tynsh:
Fixes #817 . The fix works as expected on my machine.

williamcroberts (approved):

PR #818, opened 2023-02-20
configure: detect --prefix when checking for P11_KIT

lukazlim:


configure.ac
```diff
@@ -485,6 +485,11 @@ AS_IF([test "x$have_p11kit" = "xyes"], [
 
 AM_CONDITIONAL([HAVE_P11KIT], [test "x$have_p11kit" = "xyes"])
 
+# check for cross-compilation and set makefile flag HAVE_CROSSCOMPILE if found
+#   do not change library install location if cross-compiling
+#   and install into toolchain location provided with --prefix
+AM_CONDITIONAL([HAVE_CROSSCOMPILE], [test ! -z "$(prefix)"])
```
williamcroberts, line 491: `--prefix` does not mean you're cross compiling. It means you just need a different install location.

williamcroberts (changes requested):
I'm not sure what problem your solving (better commit message please), but I think you would want another flag to control this over using `--prefix`.

configure.ac
lukazlim: Thanks for the feedback, with a clearer picture I have made changes to naming and the commit message

lukazlim:
Hi, the commit message and comments were changed with better clarity, might want to take another look at it, thanks

williamcroberts (approved):

Jakuje:
This is breaking the installation of the p11-kit modules when the prefix is set (as default in Fedora). It is not clear at all what problem is this change solving.

Jakuje:
Actually, this problem is not relevant to the p11-kit only. It fails to install the so objects to the right place too:
```
Processing files: tpm2-pkcs11-1.9.2-1.fc44.x86_64
error: File not found: /builddir/build/BUILD/tpm2-pkcs11-1.9.2-build/BUILDROOT/usr/share/p11-kit/modules/tpm2_pkcs11.module
error: Directory not found: /builddir/build/BUILD/tpm2-pkcs11-1.9.2-build/BUILDROOT/usr/lib64/pkcs11
error: File not found: /builddir/build/BUILD/tpm2-pkcs11-1.9.2-build/BUILDROOT/usr/lib64/pkcs11/libtpm2_pkcs11.so
error: File not found: /builddir/build/BUILD/tpm2-pkcs11-1.9.2-build/BUILDROOT/usr/lib64/pkcs11/libtpm2_pkcs11.so.0*
```

Jakuje:
I am not in much hurry so I am ok to wait to clarify what the reporter was trying to solve. Eventually having a new release without this would be good.

From my reading this change fab3140c27518ba45ab206ce6a58a90cf2e51c51 should have been already in the .1, but that one was building just ok in Fedora over last year[1], so it might be that some other change in .2 surfaced this issue, but I did not dig into that deeper yet.

[1] https://koji.fedoraproject.org/koji/packageinfo?packageID=28125

billatarm:
@lukazlim I need to revert this, this is causing tons of issues for others. If you would like to try again, we just need to make sure we don't break cases where --prefix is used. Probably add an extra option as suggested above.

Issue #822, opened 2023-02-27
configure --with-fapi outdated tag 


l-sousa:
The ./configure --with-fapi tag is outdated on the FAPI documentation and is present in the configure help.

williamcroberts:
> The ./configure --with-fapi tag is outdated on the FAPI documentation and is present in the configure help.

The enable/disable fapi flag during configure for the tpm2-tss project is different than the flag used by the tpm2-pkcs11 project. In tpm2-tss, it is used to enable or disable the build of the FAPI library aka libtss2-fapi. Within tpm2-pkcs11 project its used as a way to disable the FAPI backend. When including or excluding certain packages the convention is to use `--with-pkgname`, see https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/External-Software.html.

Also, downstream packagers relying on this behavior would be broken.

For those two reasons, I don't want to take this change.

PR #823, opened 2023-02-27
Fixed outdated --with-fapi option to --enable-fapi

l-sousa:
The --with-fapi tag on configure does not work. --enable-fapi is supposed to be used instead.

The configure.ac file was updated aswell as the documentation for the FAPI.

Opened an issue [here](https://github.com/tpm2-software/tpm2-pkcs11/issues/822).

This PR fixes the issue.


l-sousa:
didn't consider convention nor downstream packages! mb


thanks for the fix!

Issue #766, opened 2022-04-11
Fedora 36 ErrorCode (0x00070001) Could not create rsa key. 


traxtopel:
Following the instructions on the main page for EAP-TLS Configuration. 
Worked previously tests on Centos 8.
Any ideas to what to check? Tried removing the sqlite3 database and running tpm2_clear.

root:~# openssl req -new -engine pkcs11 -keyform engine -key "${PRIVATE_KEY};pin-value=userpin" -subj "${SUBJ}" -out client.csr
Engine "pkcs11" set.
WARNING: Listing FAPI token objects failed: "fapi:Provisioning was not executed."
Please see https://github.com/tpm2-software/tpm2-pkcs11/blob/1.7.0/docs/FAPI.md for more details
WARNING: Getting tokens from fapi backend failed.
ERROR:esys_crypto:src/tss2-esys/esys_crypto_ossl.c:754:iesys_cryptossl_pk_encrypt() ErrorCode (0x00070001) Could not create rsa key. 
ERROR:esys:src/tss2-esys/esys_iutil.c:521:iesys_compute_encrypted_salt() During encryption. ErrorCode (0x00070001) 
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:226:Esys_StartAuthSession_Async() Error in parameter encryption. ErrorCode (0x00070001) 
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:113:Esys_StartAuthSession() Error in async function ErrorCode (0x00070001) 
ERROR: Esys_StartAuthSession: esapi:Catch all for all errors not otherwise specified
ERROR: Could not start Auth Session with the TPM.
ERROR: Error unsealing wrapping key
Login failed
Login to token failed, returning NULL...
PKCS11_get_private_key returned NULL
Could not read private key from org.openssl.engine:pkcs11:pkcs11:manufacturer=STMicro;serial=0000000000000000;token=label;id=%38%32%33%66%63%62%33%30%65%63%64%32%32%64%39%32;type=private;pin-value=userpin
402C537C3B7F0000:error:03000096:digital envelope routines:fromdata_init:operation not supported for this keytype:crypto/evp/pmeth_gn.c:354:
402C537C3B7F0000:error:41800005:PKCS#11 module:ERR_CKR_error:General Error:p11_slot.c:260:
[...]

PR #833, opened 2023-03-21
PCR policy authmodel

SergiiDmytruk:
Hi.

As was mentioned in #301 about a month ago, we want to have the ability to bind use of a private key to an expected value of a PCR.

Currently the changes include a new attribute and related modifications to `tpm2_ptool`. Sending the PR just in case you can see an issue with the changes right away.

I've added `--policy` parameter to `addkey`, `import` and `link` subcommands. There is also a new subcommand called `objpol`, which is similar to `objmod` but deals only with the policy. `objpol` can be dropped if you don't see much value in it, but then it would be nice to add `--raw` to `objmod` for printing value as is rather than as a YAML.

***

Update. Fixed a bug in `tpm2_ptool` and added code for executing a policy during signing if it's present on a private key. The only callback set is "get PCR" which is enough for handling of PCR policies and their templates.

I've noticed that there seems to be an effort to maintain compatibility with `tpm2-tss` v2.0, so handling of policies is optional and is disabled if `tpm2-policy` isn't present or `--without-policy` is passed to `configure`.

There are also a couple of trivial changes in separate commits.


SergiiDmytruk:
@glance-, we ended up not needing the changes (at least not yet), so haven't touched this stuff for awhile, but some example based on notes is provided below.

#### Create a policy (`tpm-policy.json`)

Policies are described in JSON format ([specification][policy-spec]), the case of a PCR policy is quite simple:

```json
{
  "description": "PCR7 policy",
  "policy": [{
    "type": "pcr",
    "pcrs": [{
        "pcr": 7,
        "hashAlg": "sha256",
        "digest": "0000000000000000000000000000000000000000000000000000000000000000"
    }]
  }]
}
```

The `description` field is merely informational and is optional in the specification, but implementation in `tpm2-tss` requires it to be present.

Current value of a PCR can be obtained via:

```bash
tpm2_pcrread sha256:7
```

[policy-spec]: https://trustedcomputinggroup.org/wp-content/uploads/TSS_JSON_Policy_v0p7_r08_pub.pdf

#### Use it

```bash
# initialize pkcs#11
tpm2_ptool init
# create token
tpm2_ptool addtoken --pid 1 --sopin mysopin --userpin myuserpin --label tpm20
# create a key protected by the policy
tpm2_ptool addkey --label=tpm20 --key-label=mykey --userpin=myuserpin \
                  --algorithm=rsa2048 --policy="$(cat tpm-policy.json)"
```

glance-:
I think I'm missing something here. I've built a tpm2-pkcs11 with this pull-request merged ontop of master, tpm2-tss 4.0.1 and tpm2-pytss 2.2.0-rc0 which all where the latest when I started playing around with this.


When I create a key with a policy just as you've described, only changed policy from pcr 7 to pcr 8 for my tests to be simpler.

To create a simple self signed certificate with that key I've used openssl  with -provider tpm as shown in https://github.com/tpm2-software/tpm2-pkcs11/issues/766#issuecomment-1119360906 to create such a certificate:
```shell
yaml_rsa0=$(tpm2_ptool export --label=tpm20 --key-label=mykey --userpin=myuserpin)
auth_rsa0=$(echo "$yaml_rsa0" | grep "object-auth" | cut -d' ' -f2-)
openssl req -new -x509 -provider tpm2 -provider base -key "mykey".pem -passin "pass:$auth_rsa0" -subj "/CN=$HOSTNAME" -out "$HOSTNAME.pem" -days 3650 -nodes -sha256 -addext 'basicConstraints=critical,CA:FALSE' -addext 'keyUsage=digitalSignature'
```
All goes fine while:
```shell
# tpm2_pcrread sha256:8
  sha256:
    8 : 0x0000000000000000000000000000000000000000000000000000000000000000
```
I can even use it via libtpm2_pkcs11.so with ex sbsign:
```shell
sbsign --engine pkcs11 --key 'pkcs11:token=tpm20;pin-value=myuserpin' --cert "$HOSTNAME.pem" --output test-signed.efi /boot/vmlinuz-6.1.0-17-amd64
```

If I now "break" things by `tpm2_pcrextend 8:sha256=f2ca1bb6c7e907d06dafe4687e579fce76b37e4e93b7605022da52e6ccc26fd2`

I can't run above sbsign command via the pkcs11 module, but above openssl -provider tpm2 runs just fine.

How can the tpm still allow the openssl tpm provider to access to the key?

SergiiDmytruk:
Not sure, but it might be related to the fact that policy is processed only for signing [here](https://github.com/tpm2-software/tpm2-pkcs11/pull/833/commits/21a21e7541b878ad599a28e77e5dac6031ce8e08#diff-4a099f0a291619b19ee4c0b4d903b2b1e9eae10c31f2cd57825dccd372be1a44R255).

glance-:
But the whole idea is that the tpm should refuse to do anything with the key unless it knows the policy is satisfied. That means that even if I dig up the handles to the key some other way it should know that there's a policy on this key, and refuse to do anything unless it's satisfied.

I just tried the following dirty hack:
```diff
diff --git i/src/lib/sign.c w/src/lib/sign.c
index bd91a88..47a07f7 100644
--- i/src/lib/sign.c
+++ w/src/lib/sign.c
@@ -255,7 +255,8 @@ static CK_RV common_init(operation op, session_ctx *ctx, CK_MECHANISM_PTR mechan
     if (op == operation_sign) {
         rv = policy_is_satisfied(tok->tctx, tobj, tok->pobject.handle);
         if (rv != CKR_OK) {
-            return rv;
+            LOGE("policy_is_satisfied failed, continuing anyway");
+            rv = CKR_OK;
         }
     }
 
```

And now I get the libtpm2_pkcs11.so to ignore any policies and just sign things anyway, so something is definitely not the way it should be.

SergiiDmytruk:
You might want to look into something like [this key creation](https://github.com/QubesOS/qubes-antievilmaid/blob/35dd356610102f61bd953e865122a89827a7ad86/sbin/anti-evil-maid-lib-tpm2#L225-L241) and [its usage](https://github.com/QubesOS/qubes-antievilmaid/blob/35dd356610102f61bd953e865122a89827a7ad86/sbin/anti-evil-maid-lib-tpm2#L275-L281) with a policy.  This one is enforced by TPM.

glance-:
I was just expecting the key to be setup in a way requiring the policy to be fulfilled for usage, when it was created with `tpm2_ptool addkey ... --policy`

I'd guess any other user also would expect that to.

But this concludes, that this part of this feature isn't implemented in this PR, which I'd guess would be a blocker for actually merging this.

SergiiDmytruk:
> But this concludes, that this part of this feature isn't implemented in this PR, which I'd guess would be a blocker for actually merging this.

The implementation follows [this comment](https://github.com/tpm2-software/tpm2-pkcs11/issues/301#issuecomment-1433885147).  There is always a chance of misunderstanding, so maybe you're right and the implementation isn't correct.

src/lib/sign.c
```diff
@@ -149,6 +154,42 @@ static CK_RV update_pss_sig_state(token *tok, tobject *tobj) {
...
+        return CKR_OK;
+    }
+
+#ifndef HAVE_POLICY
+    UNUSED(tctx);
+    UNUSED(handle);
+    /* the policy is considered to be satisfied, but will warn about it */
+    LOGW("Found a policy, but support for enforcing it wasn't compiled in");
+    return CKR_OK;
+#else
```
williamcroberts, line 171: I would create 2 versions of this function and ifdef between the two and avoid even looking for the policy on non-policy enabled builds.

src/lib/tpm.c
```diff
@@ -4008,6 +4011,111 @@ CK_RV tpm2_getmechanisms(tpm_ctx *ctx, CK_MECHANISM_TYPE *mechanism_list, CK_ULO
...
+
+    TSS2_RC rc;
+
+    rc = Tss2_PolicySetCalcCallbacks(policy_ctx, &calc_callbacks);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOGE("Tss2_PolicySetCalcCallbacks: %s:", Tss2_RC_Decode(rc));
+        return CKR_GENERAL_ERROR;
+    }
+
+    /* XXX should we cache the session or running multiple policies is unlikely? */
```
williamcroberts, line 4089: This session handle needs to be used to authorize the object for `Esys_Sign`. You're essentially still on the password model (see my other comment in the review summary). I would modify the `tpm_ctx` struct to have two session handles:
- auth_session: This is the session that will be either the HMAC session OR the policy session from execute policy.
  - For HMAC session use, it can just live the lifetime of the token.
  - For POLICY session use, it needs to be created for each command.
      - You can use the attribute continue session and use policy reset command, but thats a performance thing that can be done later.
- enc_session: This is **important**, In the current code the HMAC session handle is also the encryption handle for protecting the traffic to and from the TPM, we need this to also be created for the lifetime of the token and passed to each command as the first unused session handle, which in most commands in session2, sometimes its session3.

williamcroberts (changes requested):
Overall great work. However, most of my comments are nits. However, this is a big one here. You need to **create** the object with the proper policy for it to work, ie set the policy hash in the `TPM2B_PUBLIC` that is fed to `Esys_Create`. Also, you need to update the object attributes to **remove** userwithauth attribute which will enforce only policy. Right now it's essentially working still on passwords.

Sorry for the delay.

erlenmayr:
What is the status of this? What is missing? This could be really useful for remote attestation of system integrity.

williamcroberts:
> What is the status of this? What is missing? This could be really useful for remote attestation of system integrity.

I don't think that author followed up with changes, but if someone wants to take over the PR, go for it.

SergiiDmytruk:
> I don't think that author followed up with changes, but if someone wants to take over the PR, go for it.

Right, we don't seem to need the feature any longer.  After seeing the review (thanks for taking a look) I thought that I might try to finish the PR anyway so the effort isn't wasted, but still didn't, so anyone actually interested in the functionality is welcome to take over.

Issue #845, opened 2023-10-05
Segmentation fault during database upgrade 7 -> 8


alitofresko:
Tested with 1.9.0, trying to use a DB created with 1.8.0.

Running a simple command with the `pkcs11-tool` we receive a segmentation fault with a specific database created with 1.8.0 lib.

Steps used to get the error:
Run `export TPM2_PKCS11_LOG_LEVEL=2` to get verbose logs
Run `pkcs11-tool --verbose --test --module /usr/lib/pkcs11/libtpm2_pkcs11.so` to perform a test
This is the output
`INFO on line: "393" in file: "/usr/src/debug/tpm2-pkcs11/1.9.0-r0/src/pkcs11.c": enter "C_GetFunctionList"
INFO on line: "393" in file: "/usr/src/debug/tpm2-pkcs11/1.9.0-r0/src/pkcs11.c": return "C_GetFunctionList" value: 0
INFO on line: "381" in file: "/usr/src/debug/tpm2-pkcs11/1.9.0-r0/src/pkcs11.c": enter "C_Initialize"
INFO on line: "41" in file: "/usr/src/debug/tpm2-pkcs11/1.9.0-r0/src/lib/backend.c": Initializing backends
INFO on line: "2626" in file: "/usr/src/debug/tpm2-pkcs11/1.9.0-r0/src/lib/db.c": Using sqlite3 DB: "/etc/tpm2_pkcs11/tpm2_pkcs11.sqlite3"
INFO on line: "2216" in file: "/usr/src/debug/tpm2-pkcs11/1.9.0-r0/src/lib/db.c": Performing DB backup at: "/etc/tpm2_pkcs11/tpm2_pkcs11.sqlite3.bak"
Segmentation fault`

After some investigation, I see that my database /etc/tpm2_pkcs11/tpm2_pkcs11.sqlite3 created with 1.8.0 has 3 rows on tobjects table, the last one has no attribute of type CKA_ALLOWED_MECHANISMS. So, the function `attr_get_attribute_by_type` returns null and the next instruction `type_from_ptr` cause a segmentation fault.

by placing an "if not null" before the function that caused the segmentation fault, the db upgrade seems to complete with success.
Attached my patch used to add more logs and the "if not null" check.
This is the relevant output of `pkcs11-tool --verbose --test --module /usr/lib/pkcs11/libtpm2_pkcs11.so` after patch applied:

[...]

PR #849, opened 2023-11-07
Skip null attribute during DB update

tracefinder:
This PR is based on a patch from https://github.com/tpm2-software/tpm2-pkcs11/issues/845 and fixes the issue when there is no CKA_ALLOWED_MECHANISMS attribute during DB migration


tracefinder:
I guess opensuse-leap has the same issue as in https://github.com/tpm2-software/tpm2-pkcs11/pull/808#issuecomment-1398563650

tracefinder:
@williamcroberts sorry for bothering you. Could you take a look at this PR?
It's not big.

idlesign:
Link to #858

williamcroberts:
Duplicate of #858

PR #866, opened 2024-06-19
Add support for importing TPM2 keys with PKCS11 vendor attributes

### LINKED ISSUE #861: Proposal for importing persistent keys and key objects into a token without relying on third-party software
Here is a proposal for importing persistent keys and key objects into a token without relying on third-party software, where it will work with both backends (FAPI/ESYSDB).

If there isn't any opposition, I'll start the implementation soon.

- Modify `struct tobject`:
  - uint32_t tpm_persistent_handle; // Persistent TPM handle
  - uint32_t tpm_esys_tr; // Loaded TPM handle

- C_GenerateKeyPair: Utilize a key template that contains vendor-specific attributes, linking the key object to either the persistent TPM key or TPM key objects.
  - For TPM key objects:
    - CKA_TPM2_PUB_BLOB
    - CKA_TPM2_PRIV_BLOB
    - CKA_TPM2_OBJAUTH // The TPM key auth value in plain text
  - For persistent TPM key:
    - CKA_TPM2_PERSISTENT_HANDLE // Allows persistent handle only
    - CKA_TPM2_OBJAUTH

- C_GenerateKeyPair:
  - If either key templates (pub/priv) include the attribute CKA_TPM2_PERSISTENT_HANDLE, indicating that a tpm_persistent_handle is used:
    - Create two tobjs: pub_tobj and priv_tobj.
    - Set pub_tobj->tpm_persistent_handle and priv_tobj->tpm_persistent_handle to the value of CKA_TPM2_PERSISTENT_HANDLE.
    - Leave the x_tobj->priv & pub fields empty.
    - Set the auth value in x_tobj using tobject_set_auth() to the wrapped value of CKA_TPM2_OBJAUTH.
    - Store the x_tobj in the backend.
    - Store the x_tobj in the global variable (token->tobjects).
  - If the pub key template includes attribute CKA_TPM2_PUB_BLOB and priv key template includes attribute CKA_TPM2_PRIV_BLOB, indicating that TPM key objects are used:
    - Create two tobjs: pub_tobj and priv_tobj.
    - Set pub_tobj->pub and priv_tobj->pub & priv to the value of CKA_TPM2_PUB_BLOB and CKA_TPM2_PRIV_BLOB, respectively.
    - Set the auth value in x_tobj using tobject_set_auth() to the wrapped value of CKA_TPM2_OBJAUTH.
    - Store the x_tobj in the backend.
    - Store the x_tobj in the global variable (token->tobjects).
  - Otherwise, follow the default implementation.

- TPM key loading using token_load_object():
  - Set tobject->tpm_esys_tr according to the rules:
    - If tobject->tpm_persistent_handle is not empty:
      - If CKA_CLASS == CKO_PRIVATE_KEY, set tobject->tpm_esys_tr to Esys_TR_FromTPMPublic(tobject->tpm_persistent_handle).
      - If CKA_CLASS == CKO_PUBLIC_KEY, set tobject->tpm_esys_tr to Esys_LoadExternal(Esys_ReadPublic(tobject->tpm_persistent_handle)).
    - Otherwise, follow the default implementation.

- C_SignInit/C_EncryptInit/C_DecryptInit Operation:
  - During the initialization operation, the tobj is loaded using token_load_object():
    - If tobj->tpm_esys_tr is already set, no action is required.
    - Otherwise:
      - If pub_tobj->tpm_persistent_handle is not empty, set tobject->tpm_esys_tr according to the above rules.
      - Otherwise, follow the default implementation to load the TPM key objects.
  - If the tobj->tpm_esys_tr is set, flush it during C_Logout -> session_ctx_logout.
    *this is the default implementation

wxleong:
- Add support for importing TPM2 keys (as persistent handle or key objects) using PKCS11 vendor-specific attributes
- Add a new CLI tool: key_import
- Add integration test
- Add docs/KEY_IMPORT_TOOL.md

Resolves #861


tools/key_import/import.c
```diff
@@ -0,0 +1,767 @@
...
+    { CKR_DOMAIN_PARAMS_INVALID, "CKR_DOMAIN_PARAMS_INVALID" },
+    { CKR_BUFFER_TOO_SMALL, "CKR_BUFFER_TOO_SMALL" },
+    { CKR_SAVED_STATE_INVALID, "CKR_SAVED_STATE_INVALID" },
+    { CKR_INFORMATION_SENSITIVE, "CKR_INFORMATION_SENSITIVE" },
+    { CKR_STATE_UNSAVEABLE, "CKR_STATE_UNSAVEABLE" },
+    { CKR_CRYPTOKI_NOT_INITIALIZED, "CKR_CRYPTOKI_NOT_INITIALIZED" },
+    { CKR_CRYPTOKI_ALREADY_INITIALIZED, "CKR_CRYPTOKI_ALREADY_INITIALIZED" },
+    { CKR_MUTEX_BAD, "CKR_MUTEX_BAD" },
+    { CKR_MUTEX_NOT_LOCKED, "CKR_MUTEX_NOT_LOCKED" },
+    { CKR_VENDOR_DEFINED, "CKR_VENDOR_DEFINED" },
```
williamcroberts, line 150: You could have just stringified these with xstr macro

tools/key_import/import.c
```diff
@@ -0,0 +1,767 @@
...
+                   CK_MECHANISM     *mech,
+                   CK_ATTRIBUTE     *pub,
+                   CK_ULONG          pub_len,
+                   CK_ATTRIBUTE     *priv,
+                   CK_ULONG          priv_len) {
+    CK_RV            rv;
+    CK_OBJECT_HANDLE pubkey;
+    CK_OBJECT_HANDLE privkey;
+
+    rv = C_GenerateKeyPair(session, mech, pub, pub_len, priv, priv_len, &pubkey, &privkey);
```
williamcroberts, line 368: I do believe this should have been C_CreateObject called twice, once for the pub key and once for the private key as this doesn't generate a key pair.

williamcroberts (commented):
@AndreasFuchsTPM and @wxleong. Am I missing something here or did you essentially reimplement tpm2_ptool link commandlet?

Also, shouldn't this be done with `C_CreateObject` and not `C_GenerateKeyPair`?

I'm looking at this as a complete revert.

wxleong:
> @AndreasFuchsTPM and @wxleong. Am I missing something here or did you essentially reimplement tpm2_ptool link commandlet?
> 
> Also, shouldn't this be done with `C_CreateObject` and not `C_GenerateKeyPair`?
> 
> I'm looking at this as a complete revert.

tpm2_ptool imports keys by making direct changes to the database and does not support the FAPI backend. The new key_import tool, however, takes a different approach to importing pre-generated TPM keys (whether persisted or as key objects). It uses newly introduced PKCS#11 vendor-specific attributes, making it compatible with any backend. Additionally, it eliminates the need to rely on Python for the operation.

As for why it was implemented under C_GenerateKeyPair instead of C_CreateObject, the decision was based on ease of use and code reuse, as the logical flow closely mirrors C_GenerateKeyPair. While it could have been implemented under C_CreateObject, this feedback unfortunately came a bit too late. Do you see this as a critical concern?

williamcroberts:
> > @AndreasFuchsTPM and @wxleong. Am I missing something here or did you essentially reimplement tpm2_ptool link commandlet?
> > Also, shouldn't this be done with `C_CreateObject` and not `C_GenerateKeyPair`?
> > I'm looking at this as a complete revert.
> 
> tpm2_ptool imports keys by making direct changes to the database and does not support the FAPI backend. The new key_import tool, however, takes a different approach to importing pre-generated TPM keys (whether persisted or as key objects). It uses newly introduced PKCS#11 vendor-specific attributes, making it compatible with any backend. Additionally, it eliminates the need to rely on Python for the operation.

Yes I see that, the overall premise is fine, but the implementation is wrong, more below.

> 
> As for why it was implemented under C_GenerateKeyPair instead of C_CreateObject, the decision was based on ease of use and code reuse, as the logical flow closely mirrors C_GenerateKeyPair. 

You don't use an API for generating key pairs as a mechanism for import as it violates the PKCS11 API. If you need to share code flows and code between two areas of a piece of software, just refactor it and place common code in functions.

> While it could have been implemented under C_CreateObject, this feedback unfortunately came a bit too late. Do you see this as a critical concern?

Critical concern as it breaks the API standard for PKCS11 and should be done under C_CreateObject. This software cannot be released. However, many of the fixes patches can stay. As of right now, I think only https://github.com/tpm2-software/tpm2-pkcs11/commit/50a636bdb19220bf926cf6664e3084a315fe4480 is being reverted. However, feel free and re-submit it under C_CreateObject.

wxleong:
Noted. i'll get this reimplemented.

PR #883, opened 2025-02-01
fix tpm2_ptool destroy command failures

Vasu77df:
The current `tpm2_ptool destroy` command does not work as intended and does not remove the persistent object

**Current behavior**:
```
❯ sudo tpm2_ptool init
action: Created
id: 1

❯ sudo tpm2_ptool destroy --pid=1
This will delete the primary object of id "1" and all associated data from db under "/root/.tpm2_pkcs11" [y/N] y
Traceback (most recent call last):
  File "/usr/bin/tpm2_ptool", line 33, in <module>
    sys.exit(load_entry_point('tpm2-pkcs11-tools==1.33.7', 'console_scripts', 'tpm2_ptool')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/tpm2_pkcs11/tpm2_ptool.py", line 26, in main
    commandlet.init('A tool for manipulating the tpm2-pkcs11 database')
  File "/usr/lib/python3/dist-packages/tpm2_pkcs11/command.py", line 102, in init
    commandlet.get()[d['which']](d)
  File "/usr/lib/python3/dist-packages/tpm2_pkcs11/commandlets_store.py", line 202, in __call__
    tr_file = bytes_to_file(pobj['handle'], d)
                            ~~~~^^^^^^^^^^
IndexError: No item with that key
```

**Updated behavior with this PR**:
```
❯ sudo tpm2_ptool destroy --pid=1
This will delete the primary object of id "1" and all associated data from db under "/root/.tpm2_pkcs11" [y/N] y
❯ echo $?
0
```

This fixes two issues:
- bytes_to_file function is replaced by get_pobject
- and we introduce a new evictcontrol method to handle removing of the persistent object
   - With the current logic `tpm2_evictcontrol` was passed an object and a handle even for a remove action.
   - We introduce a new method evictcontrol_remove to handle just removing persistent objects
     - I am not too familiar with this code, and TPM2, just in the process of learning my way around, please let me know if there's a better approach.

Thank you!

### DIFF
--- a/tools/tpm2_pkcs11/commandlets_store.py
+++ b/tools/tpm2_pkcs11/commandlets_store.py
@@ -10,7 +10,7 @@ from .command import Command
 from .command import commandlet
 
 from .db import Db
-from .utils import bytes_to_file
+from .utils import get_pobject
 from .utils import TemporaryDirectory
 from .utils import query_yes_no
 from .utils import create_primary
@@ -153,7 +153,7 @@ class InitCommand(Command):
 
                 except Exception as e:
                     if shall_evict and pobj_ctx != None:
-                        tpm2.evictcontrol(hierarchyauth, pobj_ctx)
+                        tpm2.evictcontrol_remove(hierarchyauth, pobj_ctx)
 
                     traceback.print_exc(file=sys.stdout)
                     sys.exit(e)
@@ -173,7 +173,7 @@ class DestroyCommand(Command):
         group_parser.add_argument(
             '--hierarchy-auth',
             default="",
-            help="The primary object id to remove.\n")
+            help='The authorization password for adding a primary object to the hierarchy.\n')
 
     def __call__(self, args):
         path = args['path']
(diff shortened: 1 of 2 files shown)


Vasu77df:
quick bump for a review

williamcroberts (approved):

Vasu77df:
Thank you

martinpitt:
This fix still didn't make it into Fedora 44, so here's a workaround. Maybe it's useful for anyone else:

```sh
tpm2_ptool listprimaries                 # esys-tr opens with the handle: 81000002... below
tpm2_getcap handles-persistent           # or, for an orphan with no store left to ask
tpm2_evictcontrol -C o -c 0x81000002     # see below
rm -r "$TPM2_PKCS11_STORE"
```

**Confirm the handle rather than assuming `0x810...`.** systemd persists its SRK there (`TPM2_SRK_HANDLE`, `src/shared/tpm2-util.h`), so on a systemd host that one is taken and the store's primary sits at the next free handle.

williamcroberts:
> This fix still didn't make it into Fedora 44, so here's a workaround. Maybe it's useful for anyone else:
> 
> ```shell
> tpm2_ptool listprimaries                 # esys-tr opens with the handle: 81000002... below
> tpm2_getcap handles-persistent           # or, for an orphan with no store left to ask
> tpm2_evictcontrol -C o -c 0x81000002     # see below
> rm -r "$TPM2_PKCS11_STORE"
> ```
> 
> **Confirm the handle rather than assuming `0x810...`.** systemd persists its SRK there (`TPM2_SRK_HANDLE`, `src/shared/tpm2-util.h`), so on a systemd host that one is taken and the store's primary sits at the next free handle.

Oh I'll get a PR up for Fedora.

PR #885, opened 2025-02-25
Add support for importing persistent TPM2 keys with tpm2_ptool

fdub:
- Add vendor specific attribute for persistent handle
- Add support for persistent handle in tpm2_ptool link
- Add integration tests
- Add doc section in INTEROPERABILITY.md

This PR is partially based on the changes in #866, but takes a different approach. Instead of creating a new CLI tool to import TPM2 objects using a modified C_GenerateKeyPair function, it uses tpm2_ptool to import the keys directly into the sqlite db. So it only supports the sqlite db as a backend.


williamcroberts (changes requested):
Big things:
1. After link, don't ever read the public key from the TPM again.
2. Simplify `token_load_object` logic. We shouldn't care what type of key it is, just how to ensure it's loaded for use by the tpm:
    a. for transient keys, thats an `Esys_Load()` and setting the esys_tr handle in `tpm_loadobj()`.
    b. for persistent keys, that's an `Esys_TR_FromTPMPublic()`. The public key should be added at link and just follows the existing public key logic and hit `tpm_loadobj()`.

src/lib/object.c
```diff
@@ -1299,6 +1305,16 @@ CK_RV object_init_from_attrs(tobject *tobj) {
         }
     }
 
+    a = attr_get_attribute_by_type(tobj->attrs, CKA_TPM2_PERSISTENT_HANDLE);
+    if (a && a->pValue && a->ulValueLen) {
+        CK_ULONG handle = 0;
+        if (attr_CK_ULONG(a, &handle) != CKR_OK) {
```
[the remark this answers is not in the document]
fdub: Done

src/lib/tpm.c
```diff
@@ -4685,3 +4685,52 @@ CK_RV tpm_ec_ecdh1_derive(tpm_ctx *tctx, tobject *tobj, uint8_t *ecc_point,
     return rv;
 }
 
+CK_RV tpm_get_esys_tr(
+        tpm_ctx *ctx,
+        uint32_t persistent_handle,
+        uint32_t *esys_tr,
+        uint32_t *esys_tr_pub)
+{
```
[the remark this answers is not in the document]
fdub: Done

src/lib/tpm.h
```diff
@@ -199,6 +199,29 @@ CK_RV tpm_create_transient_primary_from_template(tpm_ctx *tpm,
...
+ * @param esys_tr
+ *  An optional output (can be NULL).
+ *  The associated ESYS_TR is returned here.
+ * @param esys_tr_pub
+ *  An optional output (can be NULL).
+ *  A new key is created with only the public component from the
+ *  persistent handle. The associated ESYS_TR is returned here.
+ * @return
+ *  CKR_OK on success; otherwise, an error code.
+ */
```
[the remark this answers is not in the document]
fdub: Done

src/lib/object.c
```diff
@@ -827,10 +827,16 @@ CK_RV tobject_set_auth(tobject *tobj, twist authbin, twist wrappedauthhex) {
     return r ? CKR_OK : CKR_GENERAL_ERROR;
 }
 
-void tobject_set_handle(tobject *tobj, uint32_t handle) {
+void tobject_set_esys_tr(tobject *tobj, uint32_t esys_tr) {
     assert(tobj);
 
-    tobj->tpm_handle = handle;
+    tobj->tpm_esys_tr = esys_tr;
+}
```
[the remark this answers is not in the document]
williamcroberts: I was merely going to suggest splitting it to its own commit. If you want to propose a follow up with that, as I think it clarifies the intention better, I would gladly take it.

src/lib/tpm.c
```diff
@@ -4685,3 +4685,29 @@ CK_RV tpm_ec_ecdh1_derive(tpm_ctx *tctx, tobject *tobj, uint8_t *ecc_point,
...
+        return CKR_GENERAL_ERROR;
+    }
+
+    rval = Esys_TR_FromTPMPublic(
+            ctx->esys_ctx,
+            persistent_handle,
+            ESYS_TR_NONE,
+            ESYS_TR_NONE,
+            ESYS_TR_NONE,
+            &tr);
```
williamcroberts, line 4705: I missed this before, we want to use a session handle for session1 parameter for `Esys_TR_FromTPMPublic`. The handle to use is in `tpm_ctx `structure as field `hmac_session`, for example: `ctx->hmac_session`.

The reason that this is important, is this is where the attacker could try and swap out what the handle points to on the tpm as a way to get you use a key they supply. By using the session handle `hmac_session`, we are in an encrypted and integrity verified connection with the TPM. Think of it as ssh for the TPM.

williamcroberts (changes requested):
sorry one minor/quick change I missed in the initial review.

docs/INTEROPERABILITY.md
```diff
@@ -142,3 +142,49 @@ tpm2_ptool link --label=mytoken --userpin=myuserpin --key-label="link-key" key.p
...
+tpm2_createprimary -c token_primary.ctx
+
+# Persist the token's primary key at handle 0x81000001
+tpm2_evictcontrol -c token_primary.ctx 0x81000001
+
+# Create the token's primary key
+tpm2_createprimary -c key_primary.ctx
+
+# Persist the token's primary key at handle 0x81000002
+tpm2_evictcontrol -c key_primary.ctx 0x81000002
```
williamcroberts, line 170: Why do we create this key, as `0x81...01` is the only one referenced?

docs/INTEROPERABILITY.md
```diff
@@ -142,3 +142,49 @@ tpm2_ptool link --label=mytoken --userpin=myuserpin --key-label="link-key" key.p
...
+
+# Create a store using the token's primary
+pid="$(tpm2_ptool init --primary-handle=0x81000001 --path=~/tmp | grep id | cut -d' ' -f 2-2)"
+
+# Create a token associated with the primary object
+# Note: in production set userpin and sopin to other values.
+tpm2_ptool addtoken --pid=$pid --sopin=mysopin --userpin=myuserpin --label=mytoken --path=~/tmp
+
+# Link the key
+tpm2_ptool link --label=mytoken --userpin=myuserpin --key-label="persistent-key" 0x81000012
```
williamcroberts, line 187: missing `--path` argument to `~/tmp` where everything was set up.

williamcroberts:
@fdub I am running the sign command from your test script on IBM TPM2 Simulator from here: https://github.com/kgoldman/ibmswtpm2 and it's working fine. Perhaps this is localized to the swtpm code or is it in some other path that I need to test?

docs/INTEROPERABILITY.md
fdub: Thanks. Done.

williamcroberts:
> > @fdub I am running the sign command from your test script on IBM TPM2 Simulator from here: https://github.com/kgoldman/ibmswtpm2 and it's working fine. Perhaps this is localized to the swtpm code or is it in some other path that I need to test?
> 
> @williamcroberts For me it's working fine on a virtual TPM in a VM and on a hardware Infineon SLB9672. But not inside the testing Docker containers (neither using _swtpm_ nor _mssim_). I haven't checked running it on the IBM TPM2 Simulator outside the container, but it is what is being used with `TABRMD_TCTI=mssim`, right?

Two things:
First off, the test failure, I wonder if we need to update something in the container image. But lets not worry about that now because, secondly I missed a better solution to the problem that doesn't require the `Esys_TR_FromTPMPublic` call. When you save the persistent handle in the Python script, save the handle in the attributes as a serialized `ESYS_TR` instead of the raw `TPM2_HANDLE`. In the Python script you use the python hook for `Esys_TR_Serialize`, which is `tr_serialize` and in the C code, pkcs11 library, swap out your call `Esys_TR_FromTPMPublic` to `Esys_TR_Deserialize`.

fdub:
> > > @fdub I am running the sign command from your test script on IBM TPM2 Simulator from here: https://github.com/kgoldman/ibmswtpm2 and it's working fine. Perhaps this is localized to the swtpm code or is it in some other path that I need to test?
> > 
> > 
> > @williamcroberts For me it's working fine on a virtual TPM in a VM and on a hardware Infineon SLB9672. But not inside the testing Docker containers (neither using _swtpm_ nor _mssim_). I haven't checked running it on the IBM TPM2 Simulator outside the container, but it is what is being used with `TABRMD_TCTI=mssim`, right?
> 
> Two things: First off, the test failure, I wonder if we need to update something in the container image. But lets not worry about that now because, secondly I missed a better solution to the problem that doesn't require the `Esys_TR_FromTPMPublic` call. When you save the persistent handle in the Python script, save the handle in the attributes as a serialized `ESYS_TR` instead of the raw `TPM2_HANDLE`. In the Python script you use the python hook for `Esys_TR_Serialize`, which is `tr_serialize` and in the C code, pkcs11 library, swap out your call `Esys_TR_FromTPMPublic` to `Esys_TR_Deserialize`.

@williamcroberts Wouldn't this mean that before `tr_serialize` gets called, the ESYS_TR had to be retrieved using `tr_from_tpmpublic` in the Python script? If so, shouldn't this called be secured using a session as well?

williamcroberts:
> > > <snip/>
> > Two things: First off, the test failure, I wonder if we need to update something in the container image. But lets not worry about that now because, secondly I missed a better solution to the problem that doesn't require the `Esys_TR_FromTPMPublic` call. When you save the persistent handle in the Python script, save the handle in the attributes as a serialized `ESYS_TR` instead of the raw `TPM2_HANDLE`. In the Python script you use the python hook for `Esys_TR_Serialize`, which is `tr_serialize` and in the C code, pkcs11 library, swap out your call `Esys_TR_FromTPMPublic` to `Esys_TR_Deserialize`.
> 
> @williamcroberts Wouldn't this mean that before `tr_serialize` gets called, the ESYS_TR had to be retrieved using `tr_from_tpmpublic` in the Python script? If so, shouldn't this call be secured using a session as well?

Yes, you need to call `tr_from_tpmpublic` within the Python script first, but the security model, as documented, assumes that the system is in a good state when provisioning.

fdub:
> > > > 
> > > 
> > > 
> > > Two things: First off, the test failure, I wonder if we need to update something in the container image. But lets not worry about that now because, secondly I missed a better solution to the problem that doesn't require the `Esys_TR_FromTPMPublic` call. When you save the persistent handle in the Python script, save the handle in the attributes as a serialized `ESYS_TR` instead of the raw `TPM2_HANDLE`. In the Python script you use the python hook for `Esys_TR_Serialize`, which is `tr_serialize` and in the C code, pkcs11 library, swap out your call `Esys_TR_FromTPMPublic` to `Esys_TR_Deserialize`.
> > 
> > 
> > @williamcroberts Wouldn't this mean that before `tr_serialize` gets called, the ESYS_TR had to be retrieved using `tr_from_tpmpublic` in the Python script? If so, shouldn't this call be secured using a session as well?
> 
> Yes, you need to call `tr_from_tpmpublic` within the Python script first, but the security model, as documented, assumes that the system is in a good state when provisioning.

Alright, I‘ll get to it.

tools/tpm2_pkcs11/objects.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, objclass, attrs, auth=None, tpm_priv=None, tpm_pub=None):
     def genmechs(self, tpm2):
         raise NotImplementedError()
 
+    def force_objauth_enc(self, attrs):
+        return False
```
williamcroberts, line 50: Why do we need these? It's been a minute since I've worked in this code base, so forgive me :-p

williamcroberts (commented):
Overall looks really good, just a question

tools/tpm2_pkcs11/objects.py
fdub: Until now, in the `PKCS11Object` constructor the attribute `CKA_TPM2_OBJAUTH_ENC` was only set if `auth` and `tpm_priv` were given, so we wouldn't have set the attribute for a persistent key even if the user specified an `auth` value. I could have checked the attributes for existence of a CKA_TPM2_SERIALIZED_TR just here, but I didn't feel right to put knowledge of such specialized attribute into the most basic class of the hierarchy. Therefore this check was moved down to the `PKCS11PrivateKey` class.

PR #886, opened 2025-03-12
add SHA1-RSA-PKCS to CKM_ALLOWED_MECHANISMS

grawity:
While it's a legacy algorithm, all the other lists (ECDSA, RSA-PSS, HMAC) do include it.


williamcroberts:
Offhand this needs a test, but it might already be covered in the shell script tests. I don't see anything in the C tests.

PR #915, opened 2026-02-17
tpm: Add policy session support for signing

pothos:
For some keys on the TPM the only way to protect them from extraction
    is by policy, e.g., when they get imported and thus can't have fixedTPM
    or fixedParent set due to strict duplication validation. When a policy
    only allows signing and requires user auth it will not be possible to
    extract the key. Yet, for the signing operation now a policy session is
    required and support for that is lacking.
    Add support for creating a policy session for signing with user auth:
      policyCommandCode(TPM2_CC_Sign) + policyAuthValue
    This is saved as attribute in the SQL database by tpm2_ptool when the
    policy parameter is set. The format of the signing policy only supports
    the hardcoded value "commandcode:sign,authvalue" for now because the
    policy session generation is not dynamic but later it would be possible
    to add more specifiers and make it dynamic when the need arises.


pothos:
Updated it so that `tpm2_ptool link --policy=commandcode:sign,authvalue` sets this policy as attribute in the database. (I've deleted the env var completely.)

pothos:
Turns out this is not needed because duplication is already requiring a policy that explicitly allows extraction.

PR #916, opened 2026-02-27
PCR policy authmodel, continuation of #833

glance-:
This is work based upon https://github.com/tpm2-software/tpm2-pkcs11/pull/833 , and those commits are just rebased on top of current master.

I've tested this a bit and it looks like the policy is actually in effect now.


glance-:
I now see that the quite complicated github actions tests looks to not have libtss2-policy . If you'd like to see the test run, I can probably add that.

I've run the tests locally on my debian trixie machine, against the system libtss2-* and it works just fine. That's my indented target to run this on to, so I'm quite happy that I don't need to locally build the whole tpm2-software stack and can just build tpm2-pkcs11 for now.

glance-:
I've written a patch, replacing the shellout to tpm2_createpolicy in calculate_policy_digest with calls to tpm2_pytss TPM2_policy , but I'm uncertain in how to treat the tpm2_pytss interactions in tpm2-pkcs11. Is tpm2_pytss a optional dependency for tpm2_ptool or not?
If not, fine, we can port lots of the shellouts to tpm2-tools to tpm2_pytss calls and simplify things.

tools/tpm2_pkcs11/tpm2.py just does shellouts.
tools/tpm2_pkcs11/commandlets_keys.py and tools/tpm2_pkcs11/utils.py got tpm2_pytss added as a dependency in  042e56b004653101ea03264bde26068647ec655d


Now when thinking about it, calculate_policy_digest should probably be moved to tools/tpm2_pkcs11/tpm2.py anyhow.

pothos:
Interesting, I need to test whether this addresses my needs for `policyCommandCode(TPM2_CC_Sign) + policyAuthValue` in a more generic way than what I did in https://github.com/tpm2-software/tpm2-pkcs11/pull/915

glance-:
Feel free to. This code needs testing. https://github.com/glance-/tpm2-pkcs11/commit/4ecc139713b3c7fbdb815e50499777d50be7d7f3 contains code for more generic policy stuff rather than the current shellout to tpm2_createpolicy . I'm still thinking about replacing some/most/most of the shellouts to tpm2-tools with tpm2_pytss FFI calls, of which that is a first step.

PR #918, opened 2026-03-27
lib: ecdh1 derive: add support for CKD_SHA1_KDF key derivation function

### LINKED ISSUE #914: lib: ecdh1 derive: is missing non-null KDF
Hi,

I have been experimenting with decrypting CMS files using `openssl cms`, the openssl [pkcs11-provider](https://github.com/latchset/pkcs11-provider/) and tpm2-pkcs11, with the ultimate goal of decrypting encrypted [RAUC](https://github.com/rauc/rauc/) update bundles, which use the CMS format under the hood.

With RSA based keys this works just fine and out of the box, but trying to get prime256v1 based keys to work took me down a bit of a rabbit hole.

This GitHub issue is one third a reminder for myself on how to reproduce this, one third a hint for others that are trying something similar and one third a feature request. If I find the time I will try implementing this myself.

**TL;DR:** decrypting a CMS file that was encrypted for a recipient with a prime256v1 public key (using default openssl settings) requires us to do an ECDH key exchange using the TPM-based private key and an ephermal key stored in the CMS file (this should already work). It then requires using SHA1 as a KDF to generate a secret of arbitrary length from the fixed-length shared secret. This is not yet implemented, resulting in `CKR_MECHANISM_PARAM_INVALID` being returned from `derive`:

https://github.com/tpm2-software/tpm2-pkcs11/blob/8a3780918832a2c6ad5ca87ce4c225426704ea02/src/lib/derive.c#L196-L198

The `tpm_ec_ecdh1_derive()` function actually has support for different KDF functions and it should also be possible to perform key derivation outside the TPM since the shared secret can be read from the TPM. Implementing this may just mean forwarding the `mecha_params->kdf` to  `tpm_ec_ecdh1_derive()`.

Thank you to @ldts for laying the groundwork when it comes to ECDH support by the way.

# How to reproduce

(This assumes that you have https://github.com/tpm2-software/tpm2-pkcs11/pull/913 in your tree)

```bash
$ export PKCS11_PROVIDER_MODULE=/usr/lib/pkcs11/libtpm2_pkcs11.so.0

# Generate keypair and self-signed certificate using the prime256v1
# elliptic curve.

$ pkcs11-tool \
    --module "${PKCS11_PROVIDER_MODULE}" \
    --keypairgen --login --pin 0000 \
    --label="example-key" \
    --usage-sign --usage-decrypt --usage-derive \
    --key-type EC:prime256v1

$ openssl req \
    -new -x509 \
    -provider default -provider base -provider pkcs11 \
    -days 36500 \
    -subj "/CN=example-cert" \
    -key "pkcs11:object=example-key;pin-value=0000" \
    -out example.cert.pem

$ pkcs11-tool \
    --module "${PKCS11_PROVIDER_MODULE}" \
    --login --pin 0000 \
    --label="example-key" \
    --write-object example.cert.pem --type cert

# Encrypt message using openssl cms

$ echo "Hello World" > input.txt

$ openssl cms -encrypt \
  -in input.txt \
  -out encrypted.cms.pem \
  -outform PEM \
  example.cert.pem

# Inspect CMS file

$ openssl cms -cmsout -print -inform PEM -in encrypted.cms.pem
CMS_ContentInfo:
  contentType: pkcs7-envelopedData (1.2.840.113549.1.7.3)
  d.envelopedData:
    version: 2
    originatorInfo: <ABSENT>
    recipientInfos:
      d.kari:
        version: 3
        d.originatorKey:
          algorithm:
            algorithm: id-ecPublicKey (1.2.840.10045.2.1)
            parameter: <ABSENT>
          publicKey:  (0 unused bits)
            0000 - 04 3c 99 e3 d1 d6 6a ae-70 d7 9c 17 1d 01   .<....j.p.....
            000e - 28 d4 ee 8f c9 c7 80 d2-99 4b 42 fd 73 aa   (........KB.s.
            001c - 50 12 84 e3 0c a1 30 02-47 6f 36 ce ae 17   P.....0.Go6...
            002a - 8f 7c 8b 71 c5 96 12 a9-27 e3 99 7e 24 ce   .|.q....'..~$.
            0038 - 3d 55 90 1c c3 3e 8b c8-65                  =U...>..e
        ukm: <ABSENT>
        keyEncryptionAlgorithm:
          algorithm: dhSinglePass-stdDH-sha1kdf-scheme (1.3.133.16.840.63.0.2)
          parameter: SEQUENCE:
    0:d=0  hl=2 l=  15 cons: SEQUENCE
    2:d=1  hl=2 l=  11 prim:  OBJECT            :id-smime-alg-CMS3DESwrap
   15:d=1  hl=2 l=   0 prim:  NULL
        recipientEncryptedKeys:
            d.issuerAndSerialNumber:
              issuer: CN=example-cert
              serialNumber: 0x5FC0C5FE9943F656D5380893B39A1239AACF0632
            encryptedKey:
              0000 - 08 a3 ea 77 27 81 68 65-3b 09 f1 df 65 75   ...w'.he;...eu
              000e - 04 f2 20 30 6a d3 5b 12-be 7d a2 1c a9 d4   .. 0j.[..}....
              001c - fa 41 03 78 27 d9 fc 36-72 7d 73 ed         .A.x'..6r}s.
    encryptedContentInfo:
      contentType: pkcs7-data (1.2.840.113549.1.7.1)
      contentEncryptionAlgorithm:
        algorithm: des-ede3-cbc (1.2.840.113549.3.7)
        parameter: OCTET STRING:
          0000 - f7 1d 17 94 69 84 6c c2-                       ....i.l.
      encryptedContent:
        0000 - 89 ab 4b 4f 3e b5 74 cf-ad 9a 85 ec b2 c9 b3   ..KO>.t........
        000f - da                                             .
    unprotectedAttrs:
      <ABSENT>

# Decrypt message

$ openssl cms -decrypt \
  -in encrypted.cms.pem \
  -out output.txt \
  -inform PEM \
  -provider default -provider base -provider pkcs11 \
  -inkey 'pkcs11:object=example-key;pin-value=0000' \
  -recip example.cert.pem
Error decrypting CMS using private key
40177724967F0000:error:41000054:pkcs11:cache_key:reason(84):../pkcs11-provider-1.1.0/src/objects.c:405:Failed to cache key
```

The error message is in fact a bit confusing, but by using `TPM2_PKCS11_LOG_LEVEL=2`  we can see that `CKR_MECHANISM_PARAM_INVALID` is in fact returned.

hnez:
This PR adds support for the `CKD_SHA1_KDF` key derivation function to the Elliptic Curve Diffie Helman functionality in tpm2-pkcs11.

This Fixes https://github.com/tpm2-software/tpm2-pkcs11/issues/914 (at least partially).

Some more work is required to get it fully working:

- https://github.com/tpm2-software/tpm2-pkcs11/pull/913 should be merged as well.
- `openssl` with the `pkcs11-provider` uses a `CKS_RO_USER_FUNCTIONS` session state when calling derive. This makes `object_create()` in `src/lib/object.c` fail, because it assumes that all objects are token objects, when in fact a session object should be created here.

  See https://github.com/tpm2-software/tpm2-pkcs11/pull/921 for more information.

### DIFF
--- a/src/lib/derive.c
+++ b/src/lib/derive.c
@@ -229,10 +229,10 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
 
     /* Get the shaed secret */
     CK_BYTE* shared_secret = NULL;
-    rv = tpm_ec_ecdh1_derive(tok->tctx, tobj, /* EC private */
-                             mecha_params->public_data, /* EC point */
-                             mecha_params->public_data_len,
-                             &shared_secret, &udata.len);
+    rv = tpm_ec_ecdh_zgen(tok->tctx, tobj, /* EC private */
+                          mecha_params->public_data, /* EC point */
+                          mecha_params->public_data_len,
+                          &shared_secret, &udata.len);
     if (rv != CKR_OK) {
          tobject_user_decrement(tobj);
          return rv;
--- a/src/lib/derive.c
+++ b/src/lib/derive.c
@@ -193,10 +193,6 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
     CK_ECDH1_DERIVE_PARAMS_PTR mecha_params;
     SAFE_CAST(mechanism, mecha_params);
 
-    if (mecha_params->kdf != CKD_NULL) {
-        return CKR_MECHANISM_PARAM_INVALID;
-    }
-
     /* 2. Uncompressed EC_POINT: is a DER OCTECT string of 04||x||y */
     if (!mecha_params->public_data_len ||
         (mecha_params->public_data_len - 1) != 2 * keysize) {
@@ -227,22 +223,44 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
         return rv;
     }
 
-    /* Get the shaed secret */
-    CK_BYTE* shared_secret = NULL;
+    /* Perform an Elliptic Curve Diffie Helman key exchange using the TPM resident
+     * private key and a supplied public key.
+     * This establishes a shared secret ECC point z_point between us and the owner
+     * of the private key belonging to the supplied public key.
+     *
+     * The result can either be used as a shared secret directly (CKD_NULL) or as
+     * the input to a key derivation function, to optain a shared secret of arbitrary
+     * length (i.e. CKD_SHA1_KDF). */
+    CK_BYTE* z_point = NULL;
+    size_t z_point_len = 0;
     rv = tpm_ec_ecdh_zgen(tok->tctx, tobj, /* EC private */
                           mecha_params->public_data, /* EC point */
                           mecha_params->public_data_len,
-                          &shared_secret, &udata.len);
+                          &z_point, &z_point_len);
     if (rv != CKR_OK) {
          tobject_user_decrement(tobj);
          return rv;
     }
 
     CK_ATTRIBUTE shared_secret_attr = {
-        .ulValueLen = udata.len,
-        .pValue = shared_secret,
+        .ulValueLen = 0,
+        .pValue = NULL,
         .type = CKA_VALUE,
     };
+
+    if (mecha_params->kdf == CKD_NULL) {
+        /* Use the shared z_point as is.
+         * Without a KDF the output length depends on the ECC algorithm used
+         * and the value in the supplied template is ignored. */
+        shared_secret_attr.ulValueLen = z_point_len;
+        shared_secret_attr.pValue = z_point;
+    }
+    else {
+        free(z_point);
+        tobject_user_decrement(tobj);
+
+        return CKR_MECHANISM_PARAM_INVALID;
+    }
 
     /* Return the shared secret in a CKO_SECRET_KEY class object */
     rv = object_create(ctx, secret_template, secret_template_count, secret);
@@ -258,6 +276,6 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
     }
 
 out:
-    free(shared_secret);
+    free(shared_secret_attr.pValue);
     return rv;
 }
--- a/src/lib/derive.c
+++ b/src/lib/derive.c
@@ -8,6 +8,11 @@
 #include <openssl/ecdsa.h>
 #include <openssl/evp.h>
 
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000)
+#include <openssl/kdf.h>
+#include <openssl/core_names.h>
+#endif
+
 #include "attrs.h"
 #include "backend.h"
 #include "checks.h"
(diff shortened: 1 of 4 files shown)


test/integration/pkcs-ecdh.int.c
```diff
@@ -184,38 +185,66 @@ static void ecdh1_derive(CK_SESSION_HANDLE session, CK_BYTE id,
...
-        [0] = { .id = 0x00, .shr = { 0 }, .pub = { 0 }, .plen = 70 },
-        [1] = { .id = 0x01, .shr = { 1 }, .pub = { 1 }, .plen = 70 },
+        [0] = { .id = id_base, .shr = NULL, .pub = { 0 }, .plen = 70 },
+        [1] = { .id = id_base + 1, .shr = NULL, .pub = { 1 }, .plen = 70 },
     };
 
+    key[0].shr = malloc(shr_len);
+    key[1].shr = malloc(shr_len);
+    memset(key[0].shr, 0, shr_len);
+    memset(key[0].shr, 1, shr_len);
```
williamcroberts: Should this be `key[1]`?

test/integration/pkcs-ecdh.int.c
```diff
@@ -184,38 +185,66 @@ static void ecdh1_derive(CK_SESSION_HANDLE session, CK_BYTE id,
...
         CK_BYTE id;
     } key[] = {
-        [0] = { .id = 0x00, .shr = { 0 }, .pub = { 0 }, .plen = 70 },
-        [1] = { .id = 0x01, .shr = { 1 }, .pub = { 1 }, .plen = 70 },
+        [0] = { .id = id_base, .shr = NULL, .pub = { 0 }, .plen = 70 },
+        [1] = { .id = id_base + 1, .shr = NULL, .pub = { 1 }, .plen = 70 },
     };
 
+    key[0].shr = malloc(shr_len);
+    key[1].shr = malloc(shr_len);
```
williamcroberts: Check for malloc failures. Use the cmocka macros for null check. I just want an OOM failure to be obvious.

src/lib/derive.c
```diff
@@ -227,23 +228,86 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
...
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000)
+    else if (mecha_params->kdf == CKD_SHA1_KDF) {
+        /* Use a KDF to generate a shared secret of arbitrary length.
+         * The length is taken from the supplied template. */
+
+        shared_secret_attr.ulValueLen = udata.len;
+        shared_secret_attr.pValue = calloc(shared_secret_attr.ulValueLen,
+                                           sizeof(CK_BYTE));
+
+        EVP_KDF *kdf = EVP_KDF_fetch(NULL, OSSL_KDF_NAME_X963KDF, NULL);
```
williamcroberts, line 272: This can return NULL, so we need a check

src/lib/derive.c
```diff
@@ -227,23 +228,86 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
...
+        EVP_KDF_CTX *kctx = EVP_KDF_CTX_new(kdf);
```
williamcroberts, line 281: This could fail as well, so we need a check

src/lib/derive.c
```diff
@@ -227,23 +228,86 @@ CK_RV derive(session_ctx* ctx,  CK_MECHANISM_PTR mechanism, /* public EC point *
...
+
+        OSSL_PARAM params[4], *p = params;
```
williamcroberts: Why do I need `p` here? Can't I just do something like `parms[0] =`, `parms[1] =`. This hides the underlying storage IMO. All those code paths succeed, so `p` always gets incremented, the constants I think will make that more clear as well.

williamcroberts (changes requested):

test/integration/pkcs-ecdh.int.c
hnez: Yes, yes indeed. I have replaced the copied steps with a loop.

test/integration/pkcs-ecdh.int.c
hnez: Makes sense. I have added checks.

src/lib/derive.c
hnez: I was not to sure about which error messages to return in that case. I hope the ones I chose make some sense.

src/lib/derive.c
hnez: Ack

src/lib/derive.c
hnez: This seems to be a common pattern in code that sets up `OSSL_PARAM` arrays, which I just copied.
I guess it makes sense when setting up an array in code with a lot of conditional expressions, because it becomes difficult to keep track of the current array index in that case.
Since the control flow is rather easy here I agree that we do not have to do it this way.
I have changed the code accordingly.

PR #919, opened 2026-03-27
test: update to cmocka version 2

hnez:
The tests generate warnings / fail to build with `-Werror` on systems that ship with cmocka version 2.0 and newer because a couple of non-typesafe macros have been replaced by typesafe alternatives and the previous versions have been deprecated.

Sadly it looks like they decided against having both the new and old API in parallel (in a non-deprecation-warning-generating way) for a couple of releases and instead added and deprecated them in 2.0.

We will likely want to delay merging this PR until all relevant distributions contain cmocka 2.0 or newer.


moritzbuhl:
Some of these changes are also in #939

PR #923, opened 2026-04-09
src/lib/tpm: return NULL for twist on auth failure

petermarko:
`tpm_unseal` returns `twist` (a const char pointer alias).
Returning `false` in the error path is a type mismatch that fails with stricter compiler settings.
Return `NULL` instead.


williamcroberts (approved):
Thanks!

PR #934, opened 2026-05-21
CI Update Images

williamcroberts:
- Drop EOL images
- Add F43


src/lib/attrs.c
```diff
@@ -1332,6 +1332,7 @@ CK_RV attr_list_append_entry(attr_list **attrs, CK_ATTRIBUTE_PTR untrusted_attr)
             *attrs,
             &new_item);
     if (!x) {
+        attr_list_free(new_item);
```
hnez, line 1335: This commit looks unrelated to the other changes in the PR to me.
Should this be a separate PR instead?

The `.ci/docker.run` change in the commit also looks unrelated to the other change. Also put that in a separate commit instead?

Issue #931, opened 2026-04-30
Invalid state after deleting RSA key pair + Certificate and then creating a new RSA key pair


neo1973:
Environment: Custom PTXdist based distribution running on an Intel Atom(R) x6211E

Libraries involved:
* p11-kit - 0.23.15
* tpm2-pkcs11 - 1.9.2
* tpm2-tss - 4.1.3

I believe the problem in the title is best understood by going through the log step by step.

**Step 1: Enumerate token and objects**

Logging: [logging_1.txt](https://github.com/user-attachments/files/27263282/logging_1.txt)
 
<details>
<summary>Pretty printed token state</summary>

```
Token #1:
  initialized:  Y
  label:        nx
  manufacturer: Intel
  model:        MCC
  serial no:    0000000000000000
  PIN length:   0 - 128
  Token #1 contains 6 matching objects:
    Object #1:
      type:    public
      size:    -1
      label:
      id:      010201
    Object #2:
      type:    private
      size:    -1
      label:
      id:      010201
    Object #3:
      type:    cert
      size:    -1
      label:
      id:      010201
      subject: CN=device-122-32.devices.local
      issuer:  CN=SUB CA SOFT
      begin:   2026-04-30T22:23:54
      end:     2027-04-30T22:23:53
      hash:    24713201D2702FCE2B6C192C3D8C39CC15EC8B8B
    Object #4:
      type:    public
      size:    -1
      label:
      id:      010200
    Object #5:
      type:    private
      size:    -1
      label:
      id:      010200
    Object #6:
      type:    cert
      size:    -1
      label:
      id:      010200
      subject: CN=device-122-32.devices.local
      issuer:  CN=SUB CA SOFT
      begin:   2026-04-30T22:41:26
      end:     2027-04-30T22:41:25
      hash:    AEDD09FE1E6C17E3252D79FB3D61223994662778
Token #2:
  initialized:  N
  label:
  manufacturer: Intel
  model:        MCC
  serial no:    0000000000000000
  PIN length:   0 - 128
```
</details>

The important part is the `C_FindObject` call:
```
C_FindObjects
  IN: hSession = S72057594037927938
  IN: max_object_count = 16
INFO on line: "494" in file: "src/pkcs11.c": enter "C_FindObjects"
INFO on line: "494" in file: "src/pkcs11.c": return "C_FindObjects" value: 0
 OUT: object = (6) [ H1, H2, H3, H4, H5, H6 ]
[...]

PR #937, opened 2026-06-21
Fix in-memory object handle collision after deleting the lowest-handle object

neo1973:
Fixes the issue described in #931: In-memory object handle collision after deleting the lowest-handle objects and creating new ones (e.g. deleting and recreating a key pair). `token_add_tobject` could assign a handle that aliased an existing object until the application restarted.

Disclaimer: Claude was used to create the fix and tests.

### DIFF
--- a/src/lib/token.c
+++ b/src/lib/token.c
@@ -142,23 +142,35 @@ CK_RV token_add_tobject(token *tok, tobject *t) {
         return CKR_OK;
     }
 
-    /* minimum potential handle to add */
-    CK_OBJECT_HANDLE index = 2;
+    /*
+     * Gap before the head: the lowest handles were freed (token_rm_tobject does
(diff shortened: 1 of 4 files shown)


src/lib/token.c
```diff
@@ -142,23 +142,35 @@ CK_RV token_add_tobject(token *tok, tobject *t) {
...
+     * being assigned a handle that collided with an existing one.
+     */
+    if (tok->tobjects.head->obj_handle > 1) {
+        t->obj_handle = 1;
+        t->l.prev = NULL;
+        t->l.next = &tok->tobjects.head->l;
+        tok->tobjects.head->l.prev = &t->l;
+        tok->tobjects.head = t;
+        return CKR_OK;
+    }
```
williamcroberts, line 160: What if you removed 1, 2 from the list, this would just recreate 1 but not 2. Let me look at this deeper tomorrow

PR #939, opened 2026-08-06
Update CI images.

moritzbuhl:
* use size optimized pkcs11 images
* update to Ubuntu 26.04 from the EOL Ubuntu 18.04
* update to Fedora 44 from EOL Fedora 30

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -27,6 +27,12 @@ if [[ "$CC" == clang* ]]; then
   echo "Detecting clang, enable asan"
   config_flags="--enable-asan"
   export ASAN_ENABLED=true
+  if [[ "${DOCKER_IMAGE:-}" == "ubuntu-26.04-pkcs11" ]] ||
+    [[ "${DOCKER_IMAGE:-}" == "fedora-44-pkcs11" ]]; then
+    echo "${DOCKER_IMAGE} ships pkcs11-tool with RTLD_DEEPBIND and is incompatible with asan"
+    config_flags=""
+    export ASAN_ENABLED=false
+  fi
   echo "Exported ASAN_ENABLED=$ASAN_ENABLED"
   # To get line numbers set up the asan symbolizer
   clang_version=`$CC --version | head -n 1 | cut -d\  -f 3-3 | cut -d\. -f 1-2`
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -55,10 +55,6 @@ if [[ "$CC" == clang* ]]; then
 else
   echo "Enabling Code Coverage"
   config_flags="--disable-hardening --enable-code-coverage"
-  # Ubuntu 26.04 ships a newer cmocka that deprecates assert_in_range; keep CI unblocked.
-  if [[ "$DOCKER_IMAGE" == ubuntu-26.04* ]]; then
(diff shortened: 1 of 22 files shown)


williamcroberts:
I can take anyhelp I can get getting the CI stable again. Let me know what you need, I could propose an easier path forward too, remove all the failing CIs, and we will start adding stuff back. No reason to even test say Fedora 40 for example, it's EOL.

moritzbuhl:
Hi @williamcroberts! I recently updated and optimized the tpm2-software container images in https://github.com/tpm2-software/tpm2-software-container/pull/112 and after locally testing the CI scripts of the projects using the images I decided that I can fix the rest after publishing.

It is going as expected and my local results are not at all the same once we run them remotely via GitHub actions.

For tpm2-tss we already needed https://github.com/tpm2-software/tpm2-software-container/pull/114 and https://github.com/tpm2-software/tpm2-software-container/pull/115

The failures usually are either (1) accidentally dropped dependencies, or (2) new, improved diagnostic and debugging tooling now causing errors.

I also sneaked in some additional images in the hopes that they would (mostly) just work but I'll drop them and add them in a separate PR to keep the transition cleaner and not waste too much time on this.

Nevertheless, I will still need a few days until everything is ready.

williamcroberts:
> 
> Nevertheless, I will still need a few days until everything is ready.

Whatever you need, ping me. I'll take any help in the CI world I can.

.ci/docker.run
```diff
@@ -55,6 +55,10 @@ if [[ "$CC" == clang* ]]; then
 else
   echo "Enabling Code Coverage"
   config_flags="--disable-hardening --enable-code-coverage"
+  # Ubuntu 26.04 ships a newer cmocka that deprecates assert_in_range; keep CI unblocked.
+  if [[ "$DOCKER_IMAGE" == ubuntu-26.04* ]]; then
+    export CFLAGS="${CFLAGS:+$CFLAGS }-Wno-error=deprecated-declarations"
+  fi
```
williamcroberts: can we just fix the code to prevent this issue in a way that works for all versions of cmocka?

.ci/download-deps.sh
```diff
@@ -47,13 +43,28 @@ function get_deps() {
 
 	done;
 
-        # install tpm2-pytss package
+	# install Python helper packages. tpm2-pytss must match the tpm2-tss
+	# headers installed above, so build it from master instead of the stale
+	# PyPI wheel.
```
williamcroberts, line 48: The wheel for tpm2-pytss should be built against the system installed headers and work for any version of the library. If it doesn't we should let @whooo know so we can get eyes on this and fix it.

moritzbuhl: I did this not because of any Python specific problems. There are for example some tmp2-abrmd (even in the master branch for abrmd, working on it) and tpm2-tss related build problems with the pinned versions and newer distro's compilers.

So I decided to move all tpm2-software projects to the master branch so incompatibilities from new changes surface more quickly and (hopefully) resolve themselves too.

moritzbuhl:
@williamcroberts it finally passes. Please take a critical look!

williamcroberts (approved):

PR #943, opened 2026-08-19
ci: Add nixos-26.05-pkcs11 to the workflow matrix

JuergenReppSIT:
This PR adds NixOS 26.05 to the tpm2-pkcs11 CI workflow matrix and includes two fixes uncovered by the new CI environment:

* Quote `CC` in `AM_TESTS_ENVIRONMENT` so compiler commands containing additional flags, such as `clang -std=gnu23`, are passed correctly to the integration tests.
* Disable ASan for the NixOS PKCS#11 job because the OpenSC `pkcs11-tool` uses `RTLD_DEEPBIND`, which is incompatible with the sanitizer runtime.

With these changes, the existing build and integration tests can run successfully using the new `nixos-26.05-pkcs11` container.

PR #944, opened 2026-08-20
utils: fix fuzzer crashes

moritzbuhl:
This started as a fix for the crash in #943 but there is more.


williamcroberts (approved):

williamcroberts:
> This started as a fix for the crash in #943 but there is more.

Keep sending em up

